PR #17, opened 2022-11-01
Farm and logging

ozgunozerk:
This PR introduces:

- farm command
- logging
- better structure (for example, config related functions are moved to their own file)

also, some other simplifications to the code due to upgrades to SDK, thanks @i1i1 for the fruitful discussions


src/main.rs
```diff
@@ -1,11 +1,12 @@
...
+mod commands;
+mod config;
 mod ss58;
 mod utils;
 
 use clap::Command;
-use init::init;
+use commands::{farm::farm, init::init};
 
-fn cli() -> Command<'static> {
```
i1i1, line 8: Wow. Could you rewrite it using this `clap` tutorial:
https://docs.rs/clap/latest/clap/_derive/_tutorial/index.html

i1i1 (commented):
LGTM overall, but I would use `clap` for pasing cli commands into structures. It is fully pledged for that and is handy to use. Monorepo and substrate use that everywhere.

src/main.rs
ozgunozerk: rebase was not that successful it seems. This was removed... Will fix :)

src/commands/farm.rs
```diff
@@ -10,10 +10,13 @@ pub(crate) struct FarmingArgs {
...
-    let node: Node = Node::builder().build().await.unwrap();
+    let node: Node = Node::builder()
+        .build("node", chain_spec::gemini_2a().unwrap())
+        .await
+        .unwrap();
     match get_args_for_farming(node) {
         Ok(args) => start_farming(args).await,
-        Err(why) => println!("Error: {why}"),
+        Err(why) => panic!("Error: {why}"),
     }
```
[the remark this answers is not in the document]
ozgunozerk: Mostly yes, but then I'm losing the specific reason for this failure. `get_args_for_farming()` may fail due to various of reasons. And that information might come in handy when debugging

src/main.rs
i1i1: Nope. I meant, in general, using raw `clap` would lead to more code than using `clap::Parser`. So it would be better to switch to that. Monorepo and substrate use that everywhere.

ozgunozerk: Oh got it. I did not use it on purpose. And I followed their cookbook on git example (https://docs.rs/clap/latest/clap/_derive/_cookbook/git/index.html).
The reason is, we don't want this: `subspace -farm`
but we want this: `subspace farm`
So I'm using commands instead of arguments, that was the reason.

i1i1: I believe you can achieve that using subcommands. Take a look at this:
https://github.com/clap-rs/clap/blob/master/examples/git-derive.rs

ozgunozerk: wonderful, I'll close this PR and do a 2nd take on it then, thanks @i1i1 !

PR #23, opened 2022-11-07
Farm progress bar

### LINKED ISSUE #22: Implement a progress bar to show in the quiet mode
Related to #21

ozgunozerk:
Closes #22 

**When in progress:**
image

**When finished:**
image

@i1i1 please check the `TODO`s in the code.

### DIFF
--- a/src/commands/farm.rs
+++ b/src/commands/farm.rs
@@ -1,16 +1,12 @@
-use color_eyre::eyre::Result;
+use color_eyre::eyre::{Report, Result};
 use futures::prelude::*;
 use indicatif::{ProgressBar, ProgressStyle};
 use subspace_sdk::Farmer;
 use subspace_sdk::{chain_spec, Node, PlotDescription, PublicKey};
-
 use tracing::instrument;
 
 use crate::config::parse_config;
 use crate::utils::{install_tracing, node_directory_getter};
-
-// TODO: if there is a way to get this from the SDK, revamp this
-const SECTOR_SIZE: u64 = 2621440;
 
 #[derive(Debug)]
 pub(crate) struct FarmingArgs {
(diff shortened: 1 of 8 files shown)


ozgunozerk:
I'm not sure about the SECTOR_SIZE constant I put here, I will ask about it in the dev sync today. We should not merge till I clarify that one.

ozgunozerk:
We are good to go on my end :)

src/commands/farm.rs
```diff
@@ -42,17 +76,21 @@ async fn start_farming(farming_args: FarmingArgs) -> Result<(Farmer, Node)> {
...
-    let chain = match chain.as_str() {
+    let chain = match config_args.node_config_args.chain.as_str() {
         "gemini-2a" => chain_spec::gemini_2a().unwrap(),
         "dev" => chain_spec::dev_config().unwrap(),
         _ => unreachable!("there are no other valid chain-specs at the moment"),
     };
+    let role = match config_args.node_config_args.validator {
+        true => subspace_sdk::node::Role::Authority,
+        false => subspace_sdk::node::Role::Full,
+    };
```
i1i1, line 92: Wow, looks kinda weird. This is basically if, right?

src/commands/farm.rs
```diff
@@ -63,3 +101,27 @@ async fn prepare_farming() -> Result<FarmingArgs> {
...
+    );
+
+    while encoded < to_be_plotted {
+        encoded += 123123;
+        pb.set_position(encoded);
+        thread::sleep(Duration::from_millis(12));
+    }
+
+    pb.finish_with_message("downloaded");
+}
```
i1i1, line 122: WDUT about this dependency and using it:
https://docs.rs/zzz/latest/zzz/

src/commands/farm.rs
ozgunozerk: yeah but think about it, it is actually match. I'm not inspecting a condition, I'm matching against a value :)

ozgunozerk: making it in an if-else looked weirder to my eyes, since it made the assumption of "this is related to the business logic of the code", which is not...

src/commands/farm.rs
ozgunozerk: Hmm sounds interesting. Seamless integration with iterators is for sure nice to have. But the one I'm using is the most popular and allows more configuration (seems like). 
Don't have any strong opinions here. If you want me to switch to `zzz`, I can do that :)

i1i1: Oh, current library also has that feature:
https://github.com/console-rs/indicatif/blob/25afbed1ae264c95bf39bf88dc5b86941c65d543/examples/iterator.rs

ozgunozerk:
Will hold the PR since some stuff will change on the SDK side for plotting progress

src/commands/farm.rs
ozgunozerk: wonderful, I'll not resolve this convo to remind myself of updating this logic

i1i1 (approved):

PR #26, opened 2022-11-09
Farm info

ozgunozerk:
This PR introduces a new logic that will be helpful for `subspace info` command in the future and for the communication between `plotting_progress` and `solution` subscriptions.
Basically, this PR introduces a new file called `summary.toml`, which stores 2 things (may be more in the future):
- how many blocks have farmed till now
- is initial plotting progress finished

It solves the following problems:
- The user may want to close the app and re-open it again. And the user will still want to know how many blocks they farmed in total. 
- to be implemented `subspace info` command should not start farming, yet retrieve info about the farmer. And currently, farming starts as soon as `Farmer` is built. So, the most efficient way to retrieve info related to farmer, is again to store them in a file imo.

This PR also:
- updates rust toolchain to the newest version
- organizes `use`s in a better way
- makes `solution subscriber` to wait for `initial plotting progress` to be completed, in order to start printing to the terminal. Thus, progress bar won't be printed again and we won't see buggy behavior in the TUI

### DIFF
--- a/src/commands/wipe.rs
+++ b/src/commands/wipe.rs
@@ -14,7 +14,7 @@ pub(crate) async fn wipe() -> Result<()> {
         }
     };
     let node_directory = node_directory_getter();
-    let _ = Node::wipe(node_directory).await;
+    Node::wipe(node_directory).await?;
     println!("Node is wiped!");
 
     // TODO: modify here when supporting multi-plot
--- a/src/commands/wipe.rs
+++ b/src/commands/wipe.rs
@@ -1,4 +1,5 @@
 use color_eyre::eyre::Result;
+
 use subspace_sdk::{Node, PlotDescription};
 
 use crate::config::parse_config;
@@ -13,7 +14,7 @@ pub(crate) async fn wipe() -> Result<()> {
         }
     };
     let node_directory = node_directory_getter();
-    Node::wipe(node_directory).await?;
+    let _ = Node::wipe(node_directory).await;
     println!("Node is wiped!");
 
     // TODO: modify here when supporting multi-plot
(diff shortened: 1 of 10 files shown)


src/summary.rs
```diff
@@ -0,0 +1,95 @@
...
+    let summary_path = summary_path();
+
+    // File::create will truncate the existing file, so first
+    // check if the file exists, if not, `open` will return an error
+    // in this case, create the file
+    // if creation gives an error, that means the directory is not created
+    if File::open(&summary_path).await.is_err() {
+        if File::create(&summary_path).await.is_err() {
+            return Err(Report::msg(
+                "Your config directory is malformed. Please re-run `subspace init`",
```
achiurizo, line 33: should just keep this error message simple:

```suggestion
                "The config directory does not exist",
```

Including the location of `summary_path` would be good too.

src/summary.rs
```diff
@@ -0,0 +1,95 @@
...
+            ));
```
achiurizo, line 34: Following up from  our sync, the other approach is to just go ahead and generate that directory for the user.

As mentioned, if the user encounters some issue with `subspace init` or the other commands, we should be notifying them at that stage, and not her during the summary creation. 🙏

src/commands/wipe.rs
```diff
@@ -13,17 +14,17 @@ pub(crate) async fn wipe() -> Result<()> {
         }
     };
     let node_directory = node_directory_getter();
-    Node::wipe(node_directory).await?;
+    let _ = Node::wipe(node_directory).await;
```
i1i1, line 17: It is better to report errors to users. You can probably save result and after that return a sum of errors of those

i1i1 (approved):
LGTM

src/commands/wipe.rs
ozgunozerk: thanks, I've missed that one.

i1i1 (approved):

PR #31, opened 2022-11-10
info command implemented

### LINKED ISSUE #11: Implement a command to display info about farmer
```gherkin
Scenario: Getting information on the farmer
	Given I am farming
	When I run `subspace-cli info`
	Then I should see information on the farmer instance
```

The information can be the following:
- is a farmer instance active, if so, the status of that instance:
`SYNCING 40% - (process ID) connected to Gemini-2a`
- how much reward they gained in total  (need block explorer integration for this, let's do this later)
- how much space they pledged 
- how much space is pledged in total (by the network) - check this formula https://github.com/subspace/block-vote-authorship-collector

Considering these will be available to CLI via SDK, this should be fairly easy to implement

ozgunozerk:
This PR closes #11


src/commands/farm.rs
```diff
@@ -26,6 +26,14 @@ pub(crate) async fn farm(is_verbose: bool) -> Result<()> {
     install_tracing(is_verbose);
     color_eyre::install()?;
 
+    let instance = SingleInstance::new("subspaceFarmer")
```
jfrank-summit, line 29: should we have a constant for FARMER_INSTANCE_NAME or something along those lines? There are a couple of instances  of this.

jfrank-summit (commented):
I left a couple of comments. Also, is there a reason for the empyt "subspaceFarmer" file in the root directory?

jfrank-summit:
If this PR closes #11 shouldn't it have the following info:

> - how much reward they gained in total (need block explorer integration for this, let's do this later)
> - how much space they pledged
> - how much space is pledged in total (by the network) - check this formula https://github.com/subspace/block-vote-authorship-collector

If these aren't available from SDK yet, I wouldn't close #11 yet but maybe convert the bullet points to checkboxes and check off what has been completed.

src/commands/info.rs
```diff
@@ -0,0 +1,30 @@
+use color_eyre::{Report, Result};
+use single_instance::SingleInstance;
+
+use crate::summary::{get_farmed_block_count, get_initial_plotting_progress};
+
+pub(crate) async fn info() -> Result<()> {
+    let instance = SingleInstance::new("subspaceFarmer").map_err(Report::msg)?;
+    if !instance.is_single() {
```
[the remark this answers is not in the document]
ozgunozerk: you are right, it was an error because I thought since the application has started, there are multiple locks acquired. However, only the farmer takes the lock. My bad

src/commands/farm.rs
ozgunozerk: yes, thank you

ozgunozerk:
@jfrank-summit appreciated the detailed inspection. I'll address all of that on Monday (and maybe sooner if I have the chance in the weekend :) )

My bad on submitting this PR without testing it properly. I was unusually busy this week and this one slipped I guess. Thanks again!

src/commands/farm.rs
i1i1, line 29: Why should we have a single instance of the farmer? We supposed to be able to start 2 cli applications if they are on different chains.

src/main.rs
```diff
@@ -23,6 +23,10 @@ struct Cli {
 
 #[derive(Debug, Subcommand)]
 enum Commands {
+    #[command(
+        about = "displays info about the farmer instance (i.e. total amount of rewards, and status of initial plotting)"
+    )]
+    Info,
```
i1i1, line 29: A side note. I think we should let users to supply their own path to the config file

i1i1 (commented):
Can you also remove empty file `subspaceFarmer`?

src/main.rs
i1i1: Yeah

src/commands/info.rs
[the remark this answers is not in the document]
ozgunozerk: @jfrank-summit logic was correct. The lock was getting dropped too early, that was the bug. I've solved it

src/commands/farm.rs
ozgunozerk: this is a great suggestion. I've added a TODO in the code for that. It will be a nice to have.

src/main.rs
ozgunozerk: do you mean the config file or the summary file? If it is for summary, it's a temporary solution and will be evicted in the future as we discussed before. 
If for the config, I don't think we should let users to choose their own config path. Because then we would need another persistent storage to store this info, to be able to read the config file. And best place for that persistent storage will probably be the default config location

ozgunozerk:
I've removed the trigger for PR for now, since the new protoc dependency in workflow has granting us limited API requests, resulting in failing tests.

PR #38, opened 2022-11-15
docs

ozgunozerk:
Adds documentation

Note: `cargo doc` is currently broken. The problem is fixed, and hopefully, tomorrow we will have a new nightly release.
I don't want to merge this since I'll update the toolchain accordingly after testing `cargo doc` works fine.

However, this is not a blocker for the review process :)

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,5 +1,5 @@
 [toolchain]
-channel = "nightly-2022-11-14"
+channel = "nightly-2022-11-15"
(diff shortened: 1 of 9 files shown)


ozgunozerk:
We are good to go if we are to use `cargo doc --no-deps` one. 
If we are not to supply `--no-deps` flag, the issue is still there and we should wait for another release for the nightly @achiurizo

i1i1 (approved):

PR #49, opened 2022-11-17
remove gemini2a from the default chain value

ozgunozerk:


ozgunozerk:
will include this in node-settings PR, closing

PR #51, opened 2022-11-17
Node settings

ozgunozerk:
This pr:
- uses all the fields in `settings.toml` now in node configuration (except the `execution`, for that, waiting on SDK)
- updates nightly version, `cargo doc` command now works 100% fine
- optimizes CI, so that macOS won't occasionally fail because of `protoc`
- gets rid of `gemini-2a`

Let's try to build executables after this PR, and if it succeeds, I'll update the readme accordingly for downloading instructions

### DIFF
--- a/src/commands/farm.rs
+++ b/src/commands/farm.rs
@@ -5,12 +5,13 @@ use color_eyre::eyre::{Report, Result};
 use futures::prelude::*;
 use indicatif::{ProgressBar, ProgressStyle};
 use single_instance::SingleInstance;
+use subspace_sdk::node::{BlocksPruning, Constraints, PruningMode, RpcMethods};
 use tracing::instrument;
 
 use subspace_sdk::Farmer;
 use subspace_sdk::{chain_spec, Node, PlotDescription, PublicKey};
 
-use crate::config::parse_config;
+use crate::config::{parse_config, NodeConfig};
 use crate::summary::{create_summary_file, get_farmed_block_count, update_summary};
 use crate::utils::{install_tracing, node_directory_getter};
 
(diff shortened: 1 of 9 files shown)


src/commands/farm.rs
```diff
@@ -135,31 +136,60 @@ async fn start_farming(farming_args: FarmingArgs) -> Result<(Farmer, Node)> {
...
-    let is_validator = config_args.node_config_args.validator;
-    let role = match is_validator {
+    let role = match validator {
         true => subspace_sdk::node::Role::Authority,
         false => subspace_sdk::node::Role::Full,
     };
+    let rpc_method = match unsafe_ws_external {
+        true => RpcMethods::Unsafe,
+        false => RpcMethods::Auto,
+    };
```
i1i1, line 180: Bools are kinda undescriptive. Maybe we could just use enums which will speak for themselves in config. Like:
```toml
rpc_method = "auto"
role = "authority"
```

src/commands/farm.rs
```diff
@@ -135,31 +136,60 @@ async fn start_farming(farming_args: FarmingArgs) -> Result<(Farmer, Node)> {
...
+    let state_pruning = Some(PruningMode::Constrained(Constraints {
+        max_blocks: Some(state_pruning),
+        max_mem: None,
+    }));
+    let blocks_pruning = BlocksPruning::Some(blocks_pruning);
+    let listen_on = vec![port
+        .parse()
+        .map_err(|_| Report::msg("port in the config file is incorrect"))?];
```
i1i1, line 188: `port` here is multiaddr right? So maybe it is not a good name for it.

Also, maybe you could ask users for a vector of multiaddresses.

src/config.rs
```diff
@@ -99,7 +83,7 @@ pub(crate) fn construct_config(
             state_pruning: 1024,
             validator: true,
             name: node_name.to_owned(),
-            port: 30333,
+            port: "".to_owned(),
```
i1i1, line 94: This won't work for sure.

i1i1: You should probably don't supply any multiaddr here, but if you supply one it should be smth like `/ip4/127.0.0.1/tcp/0`.

src/commands/farm.rs
ozgunozerk: we wanted to be config file to be in its most simplest form for the user. Turning them into enums, would be better for me (and the developers), but I would argue against it for the users. Matching against a boolean is tolerable I guess for the sake of simplifying the user experience.

src/commands/farm.rs
ozgunozerk: Exactly. I was thinking of changing the name. But `Multiaddress` is not a user-friendly name I guess. It would be confusing. 
I agree on `port` is misleading, but couldn't find a better candidate tbh.

src/config.rs
ozgunozerk: Why it won't work? It worked in `dev` chain? 
If you mean it won't work on actual chains, then I can put a check in the construction, and if this is an empty string, it won't use `listen_on` method on the builder.

src/commands/farm.rs
ozgunozerk: Ok I think I found a good way that will satisfy both of us :)

src/commands/farm.rs
i1i1: you can just have it like that:
```toml
listen_on = ["/ip4/127.0.0.1/0"]
```

src/config.rs
i1i1: I mean, it shouldn't work as you just don't pass any propper value for `Multiaddr` I believe. (Or maybe `Multiaddr` here is just empty)

ozgunozerk: yes it's empty, parse function worked fine on this empty string and I was able to farm blocks on dev chain

src/commands/farm.rs
ozgunozerk: will this value be valid for each chain spec and each setting? If we may not want this hardcoded value to be present in any case, I don't want to enable it by default

i1i1 (approved):
LGTM

PR #66, opened 2022-11-29
update sdk dependency and also workflows

ozgunozerk:
- Updates the CLI according to the SDK changes
- updates workflow


i1i1 (approved):
LGTM

PR #78, opened 2022-12-06
Fix farmed blocks issue

### LINKED ISSUE #77: False positives for farmed block count
CLI probably interprets every message SDK sends for the solution stream (even for the notifications on `not found a solution`). This leads to incorrect amount of farmed blocks displayed in the application.

i1i1:
Fixes #77 

Also fixes regression for toml config serialization.

### DIFF
--- a/src/commands/farm.rs
+++ b/src/commands/farm.rs
@@ -161,21 +161,30 @@ pub(crate) async fn farm(is_verbose: bool) -> Result<(Farmer, Node, SingleInstan
 
         // solution subscriber
         tokio::spawn({
-            let farmer_clone = farmer.clone();
+            let farmer = farmer.clone();
 
-            let farmed_blocks = get_farmed_block_count()
-                .await
-                .expect("couldn't read farmed blocks count from summary");
-            let farmed_block_count = Arc::new(AtomicU64::new(farmed_blocks));
             async move {
-                for plot in farmer_clone.iter_plots().await {
+                let farmed_blocks = get_farmed_block_count()
+                    .await
+                    .expect("couldn't read farmed blocks count from summary");
+                let farmed_block_count = Arc::new(AtomicU64::new(farmed_blocks));
+                for plot in farmer.iter_plots().await {
                     plot.subscribe_new_solutions()
                         .await
-                        .for_each(|_solution| async {
-                            let total_farmed = farmed_block_count.fetch_add(1, Ordering::Relaxed);
-                            let _ = update_summary(None, Some(total_farmed)).await; // ignore the error, since we will abandon this mechanism
-                            if is_initial_progress_finished.load(Ordering::Relaxed) {
-                                println!("You have farmed {total_farmed} block(s) in total!");
+                        .for_each(|solutions| {
+                            let farmed_block_count = &farmed_block_count;
+                            let is_initial_progress_finished = &is_initial_progress_finished;
+                            async move {
+                                if !solutions.solutions.is_empty() {
+                                    let total_farmed =
+                                        farmed_block_count.fetch_add(1, Ordering::Relaxed);
+                                    let _ = update_summary(None, Some(total_farmed)).await; // ignore the error, since we will abandon this mechanism
+                                    if is_initial_progress_finished.load(Ordering::Relaxed) {
+                                        println!(
+                                            "You have farmed {total_farmed} block(s) in total!"
+                                        );
+                                    }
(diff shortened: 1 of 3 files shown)


achiurizo (approved):

src/commands/farm.rs
```diff
@@ -161,21 +161,30 @@ pub(crate) async fn farm(is_verbose: bool) -> Result<(Farmer, Node, SingleInstan
...
-            let farmer_clone = farmer.clone();
+            let farmer = farmer.clone();
 
-            let farmed_blocks = get_farmed_block_count()
-                .await
-                .expect("couldn't read farmed blocks count from summary");
-            let farmed_block_count = Arc::new(AtomicU64::new(farmed_blocks));
             async move {
-                for plot in farmer_clone.iter_plots().await {
+                let farmed_blocks = get_farmed_block_count()
```
ozgunozerk, line 167: why move this into `async move`?

src/commands/farm.rs
```diff
@@ -161,21 +161,30 @@ pub(crate) async fn farm(is_verbose: bool) -> Result<(Farmer, Node, SingleInstan
...
+                for plot in farmer.iter_plots().await {
                     plot.subscribe_new_solutions()
                         .await
-                        .for_each(|_solution| async {
-                            let total_farmed = farmed_block_count.fetch_add(1, Ordering::Relaxed);
-                            let _ = update_summary(None, Some(total_farmed)).await; // ignore the error, since we will abandon this mechanism
-                            if is_initial_progress_finished.load(Ordering::Relaxed) {
-                                println!("You have farmed {total_farmed} block(s) in total!");
+                        .for_each(|solutions| {
+                            let farmed_block_count = &farmed_block_count;
```
ozgunozerk, line 175: why these 2 extra variables and an additional `async move`?

ozgunozerk:
Thanks for the test @i1i1 , I was going to do that tomorrow but I think you just like to code :D

src/commands/farm.rs
```diff
@@ -161,21 +161,30 @@ pub(crate) async fn farm(is_verbose: bool) -> Result<(Farmer, Node, SingleInstan
...
-            let farmed_blocks = get_farmed_block_count()
-                .await
-                .expect("couldn't read farmed blocks count from summary");
-            let farmed_block_count = Arc::new(AtomicU64::new(farmed_blocks));
             async move {
-                for plot in farmer_clone.iter_plots().await {
+                let farmed_blocks = get_farmed_block_count()
+                    .await
+                    .expect("couldn't read farmed blocks count from summary");
+                let farmed_block_count = Arc::new(AtomicU64::new(farmed_blocks));
```
ozgunozerk, line 170: why move these into `async move`?

src/commands/farm.rs
```diff
@@ -161,21 +161,30 @@ pub(crate) async fn farm(is_verbose: bool) -> Result<(Farmer, Node, SingleInstan
...
                         .await
-                        .for_each(|_solution| async {
-                            let total_farmed = farmed_block_count.fetch_add(1, Ordering::Relaxed);
-                            let _ = update_summary(None, Some(total_farmed)).await; // ignore the error, since we will abandon this mechanism
-                            if is_initial_progress_finished.load(Ordering::Relaxed) {
-                                println!("You have farmed {total_farmed} block(s) in total!");
+                        .for_each(|solutions| {
+                            let farmed_block_count = &farmed_block_count;
+                            let is_initial_progress_finished = &is_initial_progress_finished;
+                            async move {
```
ozgunozerk, line 177: why 2 extra variables, and an additional `async move` ? 
Isn't it just complicating the code? Previous version was more readable imho

src/commands/farm.rs
```diff
@@ -161,21 +161,30 @@ pub(crate) async fn farm(is_verbose: bool) -> Result<(Farmer, Node, SingleInstan
...
+                                if !solutions.solutions.is_empty() {
```
ozgunozerk, line 178: i think this was the actual fix

i1i1: Yeah, right

src/commands/farm.rs
i1i1: Just to ensure that we borrow those, instead of moving

src/commands/farm.rs
i1i1: Why not though?

PR #87, opened 2022-12-13
Update sdk version and progress bars

### LINKED ISSUE #86: Loading bar on CLI is freezing, prompting users to think the application is frozen
The loading bar is freezing, prompting users to believe the application is frozen. For the time being users can refer to the logs and will see that progress is being made, but this is still sub-optimal for users and their experience.

i1i1:
Checked. Even though we removed some config fields it seem to work for current config, so no need to write migrations.

Fixes #86

New progress bars for syncing and plotting look like this:
[asciicast](https://asciinema.org/a/PwJv7vXtX5z742DZkoUIRAG0o)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3436,6 +3436,7 @@ dependencies = [
  "console",
  "number_prefix",
  "portable-atomic",
+ "unicode-segmentation",
  "unicode-width",
 ]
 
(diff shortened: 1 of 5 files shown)


ozgunozerk:
will try it myself for farming, and then approve, thanks @i1i1 !

ozgunozerk (approved):

PR #89, opened 2022-12-19
Gemini3b support

ozgunozerk:
Gemini 3b let's goo!


ozgunozerk:
Note to self, there are TOO MANY `gemini_3a/b` in the code. We should probably optimize and make it more dynamic. Changing the code everywhere for the same thing felt silly

i1i1 (approved):
Lets gooo

PR #93, opened 2022-12-21
Issue template and code of conduct

ozgunozerk:
Please review thoroughly @achiurizo @jfrank-summit .

For the code of conduct, there was this part:
```
## Enforcement

Instances of abusive, harassing, or otherwise unacceptable behavior may be reported by contacting us anonymously through [this form](https://goo.gl/forms/chVYUnA4bP70WGsL2). All complaints will be reviewed and investigated and will result in a response that is deemed necessary and appropriate to the circumstances. The project team is obligated to maintain confidentiality with regard to the reporter of an incident.

Project maintainers who do not follow or enforce the Code of Conduct in good faith may face temporary or permanent repercussions as determined by other members of the project's leadership.

If you are unsure whether an incident is a violation, or whether the space where the incident took place is covered by our Code of Conduct, **we encourage you to still report it**. We would prefer to have a few extra reports where we decide to take no action, than to leave an incident go unnoticed and unresolved that may result in an individual or group to feel like they can no longer participate in the community. Reports deemed as not a violation will also allow us to improve our Code of Conduct and processes surrounding it. If you witness a dangerous situation or someone in distress, we encourage you to report even if you are only an observer.
```

but I evicted it, since we don't have setup/link regarding an anonymous google form submission.

We can merge this as is, and then include this `Enforcement` part if you like.

@nazar-pc I tagged you, since we have a `CONTRIBUTING.md` but that does not include `CODE_OF_CONDUCT.md`, and I believe at our scale, it might be helpful.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -41,7 +41,7 @@ body:
     id: terms
     attributes:
       label: Code of Conduct
-      description: By submitting this issue, you agree to follow our [Code of Conduct](https://example.com) # TODO: fix this link
+      description: By submitting this issue, you agree to follow our [Code of Conduct](https://github.com/subspace/subspace-cli/blob/main/CODE_OF_CONDUCT.md) # TODO: fix this link
       options:
         - label: I agree to follow this project's Code of Conduct
           required: true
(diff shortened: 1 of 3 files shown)


jfrank-summit (approved):
lgtm

PR #95, opened 2022-12-29
Inline sdk configs

i1i1:
Pulling changes from https://github.com/subspace/subspace-sdk/pull/106


i1i1:
This is a breaking change, so we need to make sure that `wipe` works :D

ozgunozerk (approved):

PR #96, opened 2022-12-29
Clear unwraps

ozgunozerk:
This PR:
- gets rid of the `unwrap()`s in the code (excluding tests)
- relaxes the wipe criteria, so even when there is a breaking change, the `wipe` command will attempt to remove the plots in the default directory. And prints a nice warning message as:
```
                "could not read your config. There could have been a breaking change, or you may \
                 have an ill-formed config file. Wipe will continue... \n",
                 "However, if you have set a custom location for your plots, you will need to \
                  manually delete your plots!"
```

### DIFF
--- a/src/commands/farm.rs
+++ b/src/commands/farm.rs
@@ -64,7 +64,10 @@ pub(crate) async fn farm(is_verbose: bool) -> Result<(Farmer, Node, SingleInstan
             .map_ok(|SyncingProgress { at, target, status: _ }| (target as _, at as _))
             .map_err(|err| eyre!("Sync failed because: {err}"));
 
-        let (target_block, current_block) = syncing_progress.next().await.unwrap()?;
+        let Some(syncing_result) = syncing_progress.next().await else {
+            return Err(eyre!("backoff mechanism timed out for node syncing on the SDK side"));
+        };
+        let (target_block, current_block) = syncing_result?;
         let syncing_progress_bar = syncing_progress_bar(current_block, target_block);
 
         while let Some(stream_result) = syncing_progress.next().await {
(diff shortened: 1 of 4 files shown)


src/commands/farm.rs
```diff
@@ -202,10 +205,10 @@ fn syncing_progress_bar(current_block: u64, total_blocks: u64) -> ProgressBar {
             " {spinner:2.green} [{elapsed_precise}] {percent}% [{wide_bar:.cyan}] ({pos}/{len}) \
              {bps}, {msg}, ETA: {eta_precise} ",
         )
-        .unwrap()
-        .with_key("bps", |state: &indicatif::ProgressState, w: &mut dyn std::fmt::Write| {
-            write!(w, "{:.2}bps", state.per_sec()).unwrap()
-        })
+        .expect("hardcoded template is correct")
+        // .with_key("bps", |state: &indicatif::ProgressState, w: &mut dyn std::fmt::Write| {
```
ozgunozerk, line 209: checking this now

src/commands/farm.rs
```diff
@@ -64,7 +64,10 @@ pub(crate) async fn farm(is_verbose: bool) -> Result<(Farmer, Node, SingleInstan
             .map_ok(|SyncingProgress { at, target, status: _ }| (target as _, at as _))
             .map_err(|err| eyre!("Sync failed because: {err}"));
 
-        let (target_block, current_block) = syncing_progress.next().await.unwrap()?;
+        let Some(syncing_result) = syncing_progress.next().await else {
+            return Err(eyre!("backoff mechanism timed out for node syncing on the SDK side"));
```
i1i1, line 68: This shouldn't be an error. We can be synced and therefore sdk sends nothing and closes stream. We should just skip progress bar and go to plotting

i1i1 (changes requested):

src/commands/farm.rs
ozgunozerk: oh I see, nice then! fixing

ozgunozerk:
changes look big in the last commit, however they are plain copy pastes. I've broke down the big `farm()` function into separate other functions

i1i1 (approved):

PR #100, opened 2023-01-17
update dependencies and rust

ozgunozerk:
Updated rust nightly as in monorepo,
also fixed dependabot warnings for outdated crates


ozgunozerk:
since cargo lock changed a lot, running windows test just be to sure nothing breaks (like previously):
https://github.com/subspace/subspace-cli/actions/runs/3940025560/jobs/6740609081

i1i1 (approved):

PR #103, opened 2023-01-29
Dev chain

ozgunozerk:
adds support for devnet and dev chain


ozgunozerk:
will update this with the recent changes from the monorepo, then it will be ready

i1i1 (approved):

PR #106, opened 2023-01-30
Recursive config path creation

hazyone:
There is a problem with `subscape-cli init` if there is no `.config` path in your `$HOME`, in this case, you will get such error:
```
# ./subspace-cli init
Error: could not create the directory, because: No such file or directory (os error 2)
```
This happens because `create_dir` can't produce recursive directories creation so this command will fail.

So, there was changed:
- `create_dir` to `create_dir_all` in the `create_config` step. 
- `AlreadyExists` error check was dropped because `create_dir_all` don't raise this error.
- Added `config_path` to error logging for debug purposes

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -1,4 +1,4 @@
-use std::fs::{create_dir, File};
+use std::fs::{create_dir_all, File};
 use std::path::PathBuf;
 use std::str::FromStr;
 
@@ -157,11 +157,9 @@ pub(crate) fn create_config() -> Result<(File, PathBuf)> {
         .expect("couldn't get the default config directory!")
         .join("subspace-cli");
 
-    if let Err(err) = create_dir(&config_path) {
-        // ignore the `AlreadyExists` error
-        if err.kind() != std::io::ErrorKind::AlreadyExists {
-            return Err(eyre!("could not create the directory, because: {err}"));
-        }
+    if let Err(err) = create_dir_all(&config_path) {
+        let config_path = config_path.to_str().expect("couldn't get subspace-cli config path!");
+        return Err(eyre!("could not create the directory: `{config_path}`, because: {err}"));
     }
 
     let file = File::create(config_path.join("settings.toml"))?;


hazyone:
Hey!
> I believe we had an issue describing this somewhere, but I'm also not sure it is a good idea to create directories recursively like that. Either way we certainly want to make an error more descriptive than that.

But there are some code parts, that already using this way of path creation, like:

1. https://github.com/subspace/subspace-cli/blob/main/src/summary.rs#L49
2. https://github.com/subspace/subspace-cli/blob/main/src/utils.rs#L195

So, I've made this to make it consistent with another code

ozgunozerk:
Hi! Thanks a lot for the time spent on this. Will check today!

ozgunozerk:
Here is the related issue: https://github.com/subspace/subspace-cli/issues/82

And very similar to your suggestion: https://github.com/subspace/subspace-cli/issues/82#issuecomment-1346237583

But as the final decision, we went with leaving the config file creation responsibility to the user.
_My reasoning was the following:
- in windows and macos, the path for this config directory is already present, and more streamlined. Also, macos and windows users are more ok with default settings.
- for linux users however, a considerable amount of them likes to customize their OS with respect to their own taste. I believe that Nazar also preferred delegating the config directory creation to the user because of that reason._

However, @hazyone since you felt the need for opening that PR, I'm sure there is something to improve here.
Maybe a better error message? If you feel like there is a better way without possibly disturbing the customizations of the linux users, I'm all in!

Thanks again ^^

nazar-pc:
On Linux my concern is just about data location. Config at `.config` is fine (though it is great if can be overridden), but location of the plot must be explicit to avoid running out of space on the partition that is mounted as `/home` for instance.

hazyone:
Hey, guys!
Sorry for the long delay with the response.

Yeah, the error message is a bit confusing, because of lack of the context.
`create_config()` is the first step of the `init` stage, so you're not even gotten any output(even ASCII art, ha-ha) except:
`Error: could not create the directory, because: No such file or directory (os error 2)`. 
And the user is like: "what the heck, what kind of directory do you need?"

> Maybe a better error message? 
Maybe, but if this binary is a simple way to run blockchain node, we need to think about not-so-advanced users, that can't even run this tool without guides from somebody's blog, for example. So, they will copy-paste commands and are not ready for some other flow like: "you should create a directory with such path" :)

So, we can:
- make it explicit and do it as an additional step to the `init` phase
- leave this implicit and use recursive dir creation (like this PR)
- leave this implicit and make better logging

Next, about the plot_path.
> On Linux my concern is just about data location. Config at .config is fine (though it is great if can be overridden), but location of the plot must be explicit to avoid running out of space on the partition that is mounted as /home for instance.

It's not a problem, because due to `init` phase, user is asked to specify plot location.
```
Specify a plot location (press enter to use the default: `"/root/.local/share/subspace-cli/plots"`):
```

But then, it's a bit inconsistent, because if you provide non-existent path on this step, you will get the message:
```
supplied directory does not exist! Please enter a valid path.
```
But, if you leave this path as the suggested default, there is no existence check and this path will be recursively created when `subspace-farmer` code will be called: https://github.com/subspace/subspace/blob/main/crates/subspace-farmer/src/single_disk_plot.rs#L486

nazar-pc:
> @nazar-pc we do have a very explicit step for plot location:

Then I have no concerns about it, creating `.config` makes sense to me overall, I think other apps do create it automatically if it doesn't exist too.

ozgunozerk (approved):

PR #109, opened 2023-02-22
Update Dependencies, and small optimizations

ozgunozerk:
- updates dependencies (both sdk, and other ones)
- simplified plotting progress bar code and made it more similar to syncing progress bar for consistency
- added authority to dev setting
- now, we are listing chain options during init step, so users won't be confused:
    - because I confused myself with `dev` and `devnet` :D 
    - if I got confused, I'm sure users will too hehe

### DIFF
--- a/src/commands/farm.rs
+++ b/src/commands/farm.rs
@@ -122,7 +122,6 @@ async fn subscribe_to_node_syncing(node: Node) -> Result<()> {
             syncing_progress_bar.set_position(current_block);
             syncing_progress_bar.set_length(target_block);
         }
-
         syncing_progress_bar.finish_with_message("Syncing is done!");
     }
     Ok(())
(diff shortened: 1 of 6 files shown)


src/utils.rs
```diff
@@ -264,7 +264,7 @@ mod tests {
 
     #[test]
     fn chain_checker() {
-        assert!(ChainConfig::from_str("gemini-3c").is_ok());
+        assert!(ChainConfig::from_str("gemini3c").is_ok());
```
i1i1, line 267: This is not backwards compatible. I think we can switch to `strum` once we will remove gemini-3c and switch to 3d

src/commands/farm.rs
```diff
@@ -133,72 +133,84 @@ async fn subscribe_to_plotting_progress(
...
-        let mut plotting_progress = plot.subscribe_initial_plotting_progress().await;
-
-        if let Some(plotting_result) = plotting_progress.next().await {
-            let current_size = plotting_result.current_sector * sector_size_bytes;
-            let progress_bar = plotting_progress_bar(current_size, plot.allocated_space().as_u64());
-
-            while let Some(stream_result) = plotting_progress.next().await {
-                let current_size = stream_result.current_sector * sector_size_bytes;
-                progress_bar.set_position(current_size);
+    tokio::spawn(async move {
```
i1i1, line 136: It is better to extract this `tokio::spawn` outside of the function. You can use: `tokio::spawn(subscribe_to_plotting_progress(...))`

src/commands/farm.rs
```diff
@@ -174,45 +174,43 @@ async fn subscribe_to_solutions(
     farmer: Farmer,
     is_initial_progress_finished: Arc<AtomicBool>,
 ) {
-    tokio::spawn({
-        async move {
-            let farmed_blocks = summary
-                .get_farmed_block_count()
+    tokio::spawn(async move {
```
i1i1, line 177: Same with this one

src/utils.rs
ozgunozerk: good catch, but we won't be making a release for gemini-3c nevertheless. So, I think it's safe?

src/commands/farm.rs
ozgunozerk: much better, thanks!

src/config.rs
```diff
@@ -55,6 +57,7 @@ impl NodeConfig {
             node = node
                 .system_domain(domains::ConfigBuilder::new().core(ConfigBuilder::new().build()));
         }
+        node = node.role(Role::Authority);
```
i1i1, line 60: Can you add it to every configuration? This is also true for gemini 3c and devnet

i1i1 (approved):

PR #120, opened 2023-03-07
executor as a flag

### LINKED ISSUE #117: 🎁 [Feature Request]: Require a flag for executor instead of asking a question during `init` step

Some users are getting confused because of this. And this is an advanced option since most people will not run an executor by default.

### Code of Conduct

ozgunozerk:
Closes #117

### DIFF
--- a/src/commands/farm.rs
+++ b/src/commands/farm.rs
@@ -50,7 +50,7 @@ pub(crate) async fn farm(is_verbose: bool) -> Result<(Farmer, Node, SingleInstan
         ChainConfig::Dev =>
             chain_spec::dev_config().expect("cannot extract the dev chain spec from SDK"),
         ChainConfig::DevNet =>
-            chain_spec::devnet_config().expect("cannot extrat the devnet chain spec from SDK"),
+            chain_spec::devnet_config().expect("cannot extract the devnet chain spec from SDK"),
     };
 
     let node = node_config
--- a/src/commands/init.rs
+++ b/src/commands/init.rs
@@ -11,20 +11,20 @@ use crate::config::{
 use crate::utils::{
     cache_directory_getter, get_user_input, node_directory_getter, node_name_parser,
     plot_directory_getter, plot_directory_parser, print_ascii_art, print_version,
-    reward_address_parser, size_parser, yes_or_no_parser,
+    reward_address_parser, size_parser,
 };
 
 /// implementation of the `init` command
 ///
 /// prints a very cool ascii art,
 /// creates a config file from the user inputs
-pub(crate) fn init() -> Result<()> {
+pub(crate) fn init(executor: bool) -> Result<()> {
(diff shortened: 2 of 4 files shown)


i1i1:
I think we want different behavior. I think overall workflow should be like this:
```sh
$ subspace init
Enter your farmer/reward address: ...
Enter your node name to be identified on the network: ...
Specify a plot location: ...
Specify a plot size: ...
Specify the chain to farm: ...
Configuration has been generated at /home/i1i1/.config/subspace-cli
Ready for lift off! Run the follow command to begin:
`subspace farm`
$ subspace farm --executor
...
```

It would create way better ergonomics for users and would actually make sense, because they won't need to regenerate config in order to run an executor.

But for achieving that, it would make sense to change `config.toml` layout (so that it won't be reused with SDK). For example, in the following way:
```toml
# regular node options (CLI specific)
[node]
chain = "gemini-3d"
path = "./path-to-node-data"
name = ",,,"

[node.extra-options]
# any kind of option user wants to add and override for SDK and therefore Substrate

[node.extra-executor-options]
# any kind of option user wants to add and override for SDK and therefore Substrate

[[farmer.plot]]
path = "..."
plot-size = "..."

[farmer.extra-options]
# any kind of option user wants to add and override for SDK
```

WDUT?

ozgunozerk:
This was my initial thought as well, but I didn't like the idea, because:
- farm function calls `validate config` function, which creates a node struct. This struct is not editable after constructed (because of SDK API, and I think it should stay as is)
- farm function should stay pure imo, and do not touch the config. Being an executor is related to config, which is `init` function
- in order to farm function to achieve what you have described above, it should basically do the following:
    - read config, and extract the values from the node struct (or do not construct the node struct in the first place, and read the config in a different way)
    - check for the executor flag, change the config accordingly
    - then read the config, and construct the node

I can do it, but feels like an anti-pattern.


Pro Reasons for `executor` flag in `init` step are:
- first and foremost, this is an advanced setting, and we expect advanced users to use it. So, this is not the most common use case
- advanced users can as well edit the config file, but we are just providing them some convenience here imo
- being able to switch the `executor` flag all the time, I don't see any reason for that. 
- providing the `--executor` flag in the `init` or `farm` step is the same effort from the users perspective
- functions are reamining pure (farm is not going to mess with config, only read it)
- no weird connections/hacks needs to be done to the `farm` function
- SDK API does not need to change

We can discuss it for sure, but these were the reasons I did not follow your suggestion

i1i1 (commented):
Okay, everything makes sense. We don't want users to be confused with executor option, so we push it to the CLI flag during init (as executors don't change to farmers often).

ozgunozerk:
However, your suggestion to simplify the config was right on point. I can see know how it can be done. I will close this PR and open a new one with the simplified config. Thanks @i1i1

Issue #129, opened 2023-03-15
🐞 [Bug]: generate the correct command suggestion after the `init` step


ozgunozerk:
### What happened?

after `init`, the following message is displayed:
---
```
Run the following command to being:
`./subspac-cli farm`
```
---

### Version

v0.1.8

### platform

- [X] linux
- [X] windows
- [X] macos

### Expected behavior

the application should dynamically generate the executable name, and craft the command according to each operating system

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #135, opened 2023-03-16
Small fixes

### LINKED ISSUE #134: 🎁 [Feature Request]: if user name is `root`, don't accept it

CLI takes the username from the OS. `root` is a very common username, and it will not be a good choice for users to distinguish themselves in Telemetry.

### Code of Conduct

ozgunozerk:
Closes #134 and #129


src/utils.rs
```diff
@@ -80,7 +97,8 @@ where
 
 /// node name should be ascii, and should begin/end with whitespace
 pub(crate) fn node_name_parser(node_name: &str) -> Result<String> {
-    if !node_name.trim().is_empty() {
+    let node_name = node_name.trim();
+    if !node_name.is_empty() && node_name != "root" {
```
i1i1: It would be better to set default value to `None` if the username is `root`

i1i1 (approved):

PR #139, opened 2023-03-17
Update sdk

ozgunozerk:
when syncing is completed, now we mention this is only for `initial syncing`, and regular syncing is continuing in the background. Some of the users were confused by it.


i1i1:
CI Fails :(

i1i1 (approved):

PR #147, opened 2023-03-21
Version command

### LINKED ISSUE #146: 🎁 [Feature Request]: Add command to get current version

The ability to somehow determine the version of an individual executable.

### Code of Conduct

ozgunozerk:
This PR:
- adds `-V` and `--version` flags so users can get the version even if they change the name of the executable
- gets the executable name from `std::env`
- crafts the `subspace-cli farm` command dynamically w.r.t the current directory (if current directory is where the executable is located, only `./subspace-cli farm` is generated. If not, `/path/to/executable/subspace-cli farm` is generated

@ImmaZoni tagging you, so that you can change the docs accordingly when we have a new release

closes #146

### DIFF
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -49,17 +49,9 @@ pub(crate) fn generate_run_executable_command() -> Result<String> {
     // if current dir is equal to the dir of the exec, just recommend the filename
     // else, recommend the full path
     let exec_name = if current_dir.eq(exec_dir) {
-        format!(
-            "./{}",
-            exec_full_path
-                .file_name()
-                .expect("filename cannot be empty")
-                .to_str()
-                .expect("conversion cannot be empty for filenames")
-                .to_owned()
-        )
+        format!("./{:?}", exec_full_path.file_name().expect("filename cannot be empty"))
     } else {
-        exec_full_path.to_str().expect("conversion cannot be empty for filenames").to_owned()
+        format!("{exec_full_path:?}")
     };
 
     Ok(format!("`{exec_name} farm`"))
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -41,20 +41,10 @@ pub(crate) fn print_version() {
     println!("version: {version}");
 }
 
-pub(crate) fn generate_run_executable_command() -> Result<String> {
-    let exec_full_path = std::env::current_exe()?;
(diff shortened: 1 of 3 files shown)


src/utils.rs
```diff
@@ -41,20 +41,14 @@ pub(crate) fn print_version() {
...
-    let command = format!("`./{executable_name} farm`");
-
-    println!("{command}");
+pub(crate) fn generate_run_executable_command() -> Result<String> {
+    let executable_name = std::env::current_exe()?
+        .file_name()
+        .expect("filename cannot be empty if `current_exe()` returned ok")
+        .to_str()
+        .expect("osStr to str conversion cannot be empty for filenames")
+        .to_owned();
```
i1i1: I think it can be not `utf-8` though

i1i1: Yes, I do. Because if someone will use Cyrillic symbols on Windows, it will panic

src/utils.rs
```diff
@@ -41,20 +41,20 @@ pub(crate) fn print_version() {
...
+    // if current dir is equal to the dir of the exec, just recommend the filename
+    // else, recommend the full path
+    let exec_name = if current_dir.eq(exec_dir) {
+        format!("./{:?}", exec_full_path.file_name().expect("filename cannot be empty"))
+    } else {
+        format!("{exec_full_path:?}")
+    };
 
-    println!("{command}");
+    Ok(format!("`{exec_name} farm`"))
```
i1i1, line 57: I think you can just use `std::env::current_exe()` instead of this function

ozgunozerk: If the user runs the executable from another directory, the suggested command wont work via copy paste. This should be better for users, and wasnt a big hassle to me. I think it is useful

ozgunozerk: This way, it will always work via simple copy paste without any change required from the end-user

ozgunozerk: While in the same directory with executable, i didnt want to print a huge line a command including all the path. And printing only the name of the executable wasnt cutting it when the directory was not the same

i1i1: Okay, how about:
```rust
PathBuf::from(std::env::args().next().expect("Command always exists"))
```

i1i1 (approved):

PR #157, opened 2023-03-27
fix farmer silent exit

### LINKED ISSUE #152: 🐞 [Bug]: Farmer quietly exits
### What happened?

More context there:
https://discord.com/channels/864285291518361610/1088790894196568126

### Version

0.1.11

### platform

- [X] linux
- [ ] windows
- [ ] macos

### Expected behavior

Plotting works

### Code of Conduct

ozgunozerk:
closes #152

### DIFF
--- a/src/commands/farm.rs
+++ b/src/commands/farm.rs
@@ -88,16 +88,15 @@ pub(crate) async fn farm(is_verbose: bool, executor: bool) -> Result<()> {
             is_initial_progress_finished.clone(),
         ));
 
-        wait_on_farmer(Some(plotting_sub_handle), Some(solution_sub_handle), farmer, node).await?;
+        wait_on_farmer(Some((plotting_sub_handle, solution_sub_handle)), farmer, node).await?;
     } else {
-        wait_on_farmer(None, None, farmer, node).await?;
+        wait_on_farmer(None, farmer, node).await?;
     }
     Ok(())
 }
 
 async fn wait_on_farmer(
-    maybe_plotting_handle: Option<JoinHandle<()>>,
-    maybe_solution_handle: Option<JoinHandle<()>>,
+    maybe_handles: Option<(JoinHandle<()>, JoinHandle<()>)>,
     farmer: Arc<Farmer>,
     node: Node,
 ) -> Result<()> {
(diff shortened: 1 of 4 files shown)


i1i1 (commented):
It looks weird. I think you just want to revert this change and make it work like this:
```rust
if is_verbose {
   signal::ctrl_c().await?;
} else {
   ....
}
```

ozgunozerk:
No I didn't. 

In the `non-verbose` mode, we are waiting on the subscriptions.
Silent exit bug is happening on the `verbose` mode, since there is nothing being awaited on. 

So, as a psuedo-code:
```
if non-verbose:
    initialize and handle subscriptions

when kill signal received:
    if non-verbose:
        close subscriptions
    close farmer
    close node
```

i1i1:
You right, makes sense. I would just prefer inlining those and making 2 different code paths. For example, this way, you won't even need to wrap `Farmer` with `Arc`.

ozgunozerk:
`Arc` is not introduced with this PR. 
`Arc` is required because now `Farmer` is no longer `Clone`. And subscriptions do require `Farmer` as the argument. 
`Arc` was a simple enough way to get around this, while protecting safety.

ozgunozerk:
that doesn't work since line 109 and 110 -> `plotting_subscribe_handle` and `solution_subscribe_handle` are not found. 
You are defining them in the `else` condition, yet trying to use them out of `else`

src/commands/farm.rs
```diff
@@ -74,52 +75,71 @@ pub(crate) async fn farm(is_verbose: bool, executor: bool) -> Result<()> {
...
-        let handle = tokio::spawn(async move {
-            // if one of the subscriptions have not aborted yet, wait
-            // Plotting might end, so we ignore result here
-            let _ = plotting_subscribe_handle.await;
-            solution_subscribe_handle.await.expect_err("Solution subscription never ends");
+    // closing the subscriptions if there are any
+    if maybe_solution_handle.is_some() && maybe_plotting_handle.is_some() {
+        maybe_plotting_handle.as_ref().expect("check is above").abort();
+        maybe_solution_handle.as_ref().expect("check is above").abort();
+    }
```
i1i1, line 116:
```suggestion
    if let Some((plotting_handle, solution_handle)) = maybe_handles.as_ref() {
        plotting_handle.abort();
        solution_handle.abort();
    }
```

i1i1 (approved):

PR #159, opened 2023-03-29
Fix macos signing

ozgunozerk:
After this, we will have proper macos signing. I did test it via manual dispatch (downloaded the executables, everything is fine 🥳 )


ImmaZoni (approved):
Looks good to me, only minor feedback would be changing the verbage/filenames in the workflow to reflect the fact its signing cli not binaries, but thats more of a nice to have....

i1i1 (approved):

PR #164, opened 2023-04-06
Subscription error propagating

### LINKED ISSUE #163: 🐞 [Bug]: Inspect reward events and revamp the logic for `# farmed blocks`

Right now, `farmed blocks count` refers to every solution found by the farmer. However, finding a solution does not necessarily mean this very solution will be used for the farming of the block. In other words, there can be multiple solutions for a block, and another solution can be picked as the final solution. 

The end-user probably does only care about `how many blocks they farmed`, not `how many solutions they have found`. 

Will update CLI code accordingly once SDK can expose `reward` events

### Code of Conduct

ozgunozerk:
closes #163 

This PR:
- adds error propagation to subscriptions (tokio tasks)
- overhauls summary, so that we are now able to track:
    - vote count
    - rewards earned
    - blocks farmed

Because of `summary` changes, this will be a breaking change.
I'll implement the logic for wiping summary in the next pr, since i will make more changes to `wipe` and `info`

### DIFF
--- a/src/commands/farm.rs
+++ b/src/commands/farm.rs
@@ -1,5 +1,5 @@
 use std::io::Write;
-use std::sync::atomic::{AtomicBool, AtomicU64, Ordering};
+use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Arc;
 
 use color_eyre::eyre::{eyre, Context, Report, Result};
@@ -19,6 +19,7 @@ use crate::utils::{install_tracing, raise_fd_limit};
 
 /// allows us to detect multiple instances of the farmer and act on it
 pub(crate) const SINGLE_INSTANCE: &str = ".subspaceFarmer";
+type MaybeHandles = Option<(JoinHandle<Result<()>>, JoinHandle<Result<()>>)>;
 
 /// implementation of the `farm` command
 ///
@@ -44,6 +45,7 @@ pub(crate) async fn farm(is_verbose: bool, executor: bool) -> Result<()> {
     raise_fd_limit();
 
     let Config { chain, farmer: farmer_config, node: mut node_config } = validate_config()?;
+    let reward_address = farmer_config.reward_address;
 
     // apply advanced options (flags)
     if executor {
@@ -52,8 +54,9 @@ pub(crate) async fn farm(is_verbose: bool, executor: bool) -> Result<()> {
     }
 
     println!("Starting node ...");
-    let node =
-        node_config.build(chain.clone(), is_verbose).await.context("error building the node")?;
+    let node = Arc::new(
+        node_config.build(chain.clone(), is_verbose).await.context("error building the node")?,
+    );
     println!("Node started successfully!");
 
     if !matches!(chain, ChainConfig::Dev) {
@@ -67,8 +70,7 @@ pub(crate) async fn farm(is_verbose: bool, executor: bool) -> Result<()> {
     let summary = Summary::new(Some(farmer_config.plot_size)).await?;
 
     println!("Starting farmer ...");
-    let farmer = farmer_config.build(&node).await?;
-    let farmer = Arc::new(farmer);
+    let farmer = Arc::new(farmer_config.build(&node).await?);
     println!("Farmer started successfully!");
 
     let maybe_handles = if !is_verbose {
@@ -85,9 +87,9 @@ pub(crate) async fn farm(is_verbose: bool, executor: bool) -> Result<()> {
 
         let solution_sub_handle = tokio::spawn(subscribe_to_solutions(
             summary.clone(),
-            farmer.clone(),
+            node.clone(),
             is_initial_progress_finished.clone(),
-            farmer_config.reward_address,
+            reward_address,
         ));
 
         Some((plotting_sub_handle, solution_sub_handle))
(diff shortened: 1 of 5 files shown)


src/commands/farm.rs
```diff
@@ -101,26 +105,41 @@ pub(crate) async fn farm(is_verbose: bool, executor: bool) -> Result<()> {
...
+            res = solution_handle => {
+                // first level is join handle, second layer is actual result from the task
+                match res {
+                    Ok(Ok(())) => unreachable!("solution subscription never ends"),
+                    Ok(Err(err)) => return Err(eyre!("solution subscription crashed with err: {err}")),
+                    Err(err) => return Err(eyre!("couldn't join subscription handle with err: {err:?}"))
+                }
+            }
+            // cannot inspect plotting subscription for errors, since it may end and quit from
+            // `tokio::select`
```
ozgunozerk, line 134: @i1i1 
i'm not a big fan of this code, but couldn't find a better solution. If you have any ideas on refactoring here, I'm all ears!

src/commands/farm.rs
```diff
@@ -101,26 +105,41 @@ pub(crate) async fn farm(is_verbose: bool, executor: bool) -> Result<()> {
 
 #[instrument]
 async fn wait_on_farmer(
-    maybe_handles: Option<(JoinHandle<()>, JoinHandle<()>)>,
+    mut maybe_handles: MaybeHandles,
     farmer: Arc<Farmer>,
```
i1i1, line 109: I think you no longer need an `Arc` for a farmer

src/commands/farm.rs
```diff
@@ -101,26 +105,41 @@ pub(crate) async fn farm(is_verbose: bool, executor: bool) -> Result<()> {
...
-    println!(
-        "\nWill try to gracefully exit the application now. If you press ctrl+c again, it will \
-         try to forcefully close the app!"
-    );
-
-    // closing the subscriptions if there are any
-    if let Some((plotting_handle, solution_handle)) = maybe_handles.as_ref() {
-        plotting_handle.abort();
-        solution_handle.abort();
+    if let Some((_plotting_handle, solution_handle)) = maybe_handles.as_mut() {
```
i1i1, line 114:
```suggestion
    if let Some((plotting_handle, solution_handle)) = maybe_handles.as_mut() {
```

src/commands/farm.rs
```diff
@@ -101,26 +105,41 @@ pub(crate) async fn farm(is_verbose: bool, executor: bool) -> Result<()> {
...
+        tokio::select! {
+            _ = signal::ctrl_c() => {
+               println!(
+                    "\nWill try to gracefully exit the application now. Please wait for a couple of seconds... If you press ctrl+c again, it will \
+                    try to forcefully close the app!"
+                );
+                let (plotting_handle, solution_handle) = maybe_handles.as_ref().expect("check is up there");
```
i1i1, line 121:
```suggestion
```

src/commands/farm.rs
```diff
@@ -101,26 +105,41 @@ pub(crate) async fn farm(is_verbose: bool, executor: bool) -> Result<()> {
...
+                plotting_handle.abort();
+                solution_handle.abort();
+            }
+            res = solution_handle => {
+                // first level is join handle, second layer is actual result from the task
+                match res {
+                    Ok(Ok(())) => unreachable!("solution subscription never ends"),
+                    Ok(Err(err)) => return Err(eyre!("solution subscription crashed with err: {err}")),
+                    Err(err) => return Err(eyre!("couldn't join subscription handle with err: {err:?}"))
+                }
```
i1i1, line 131:
```suggestion
                return res.context("couldn't join subscription handle")?.context("solution subscription crashed");
```

src/commands/farm.rs
```diff
@@ -205,46 +229,91 @@ async fn subscribe_to_plotting_progress(
...
+                .await
+                .context("couldn't read farmed block count value, summary was corrupted")?,
+            summary
+                .get_vote_count()
+                .await
+                .context("couldn't read vote count value, summary was corrupted")?,
+            summary
+                .get_total_rewards()
+                .await
+                .context("couldn't read total rewards value, summary was corrupted")?,
```
i1i1, line 299: I don't think you need to parse the file 3 times here. Summary is the smart wrapper around file, right? I think it might be worth redoing it. Maybe even removing it, as you can scan events from all blocks which are available to get all those stats

ozgunozerk: summary logic is to be abandoned, so that's why i'm ok with hacky solutions and don't wanna invest much time. 
+1 for removing it, let us discuss how to query blocks to deduce `total rewards` and so on... I can do that in a later PR

src/commands/farm.rs
ozgunozerk: This is much better!

src/commands/farm.rs
ozgunozerk: I still do unfortunately. 
Plotting subscription needs `farmer`.
`farmer` being not `Clone`, it tries to move `Farmer`, and I also need to close `Farmer` after this subscription

src/commands/farm.rs
ozgunozerk: I'm satisfied after that great suggestion.

src/commands/farm.rs
```diff
@@ -101,26 +105,41 @@ pub(crate) async fn farm(is_verbose: bool, executor: bool) -> Result<()> {
...
+                match res {
+                    Ok(Ok(())) => unreachable!("solution subscription never ends"),
+                    Ok(Err(err)) => return Err(eyre!("solution subscription crashed with err: {err}")),
+                    Err(err) => return Err(eyre!("couldn't join subscription handle with err: {err:?}"))
+                }
+            }
+            // cannot inspect plotting subscription for errors, since it may end and quit from
+            // `tokio::select`
+        }
+    } else {
```
i1i1: This one works:
```suggestion
        if let Some((plotting_handle, solution_handle)) = maybe_handles.as_mut() {
            futures::select! {
            _ = signal::ctrl_c().fuse() => {
               println!(
                    "\nWill try to gracefully exit the application now. Please wait for a couple of seconds... If you press ctrl+c again, it will \
                    try to forcefully close the app!"
                );
                plotting_handle.abort();
                solution_handle.abort();
            }
            res = solution_handle.fuse() => {
                return res.context("couldn't join subscription handle")?.context("solution subscription crashed");
            }
        }
    } else {
```

src/commands/farm.rs
i1i1: I would better remove it right now if it will be abandoned. There is no need to make even more logic around it.

i1i1 (approved):

PR #165, opened 2023-04-06
Wipe info revamp

ozgunozerk:
This PR:
- revamps `wipe` and `info` command w.r.t `summary` changes
- adds tokio console tracing

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -10,7 +10,7 @@ mod summary;
 mod utils;
 
 use clap::{Parser, Subcommand};
-use color_eyre::eyre::Report;
+use color_eyre::eyre::{Context, Report};
 use color_eyre::Help;
 use tracing::instrument;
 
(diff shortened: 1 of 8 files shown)


src/main.rs
```diff
@@ -56,12 +56,7 @@ enum Commands {
         executor: bool,
     },
     #[command(about = "wipes the node and farm instance (along with your plots)")]
-    Wipe {
-        #[arg(long, action)]
-        farmer: bool,
-        #[arg(long, action)]
-        node: bool,
```
i1i1, line 63: I think you shouldn't remove flags. I think the path should be as follow:
```
if terminal && !farmer && !node {
  interactive_prompts();
} else {
  current_code();
}
```

ozgunozerk: will think about that, thanks!

i1i1: `terminal` bool here should be check whether stdout is terminal. This crate will do the job I think: https://crates.io/crates/is-terminal

ozgunozerk:
I'll spawn tokio tasks with names in the future. I might have other stuff for optimizing the debugging, so saving it for a later pr

ozgunozerk:
closed in favor of #166 
i don't want to edit cargo.lock file manually :)

PR #169, opened 2023-04-12
Update sdk

### LINKED ISSUE #170: 🐞 [Bug]: Do not give error for `wipe` if `summary` is not found
### What happened?

When `wipe` command tries to delete everything, and if `summary` is missing, it crashes with an error message. 

### Version

0.3.0

### platform

- [X] linux
- [X] windows
- [X] macos

### Expected behavior

Instead of crash, it could simply print this out

### Code of Conduct

ozgunozerk:
This ps:
- gets rid of `unstable_tokio` for default use case
- separates tests into its own file (since util.rs started to get fat)
- updates sdk revision
- refactoring `summary` structs 
- closes #170


.github/workflows/release.yml
```diff
@@ -56,7 +56,7 @@ jobs:
     runs-on: ${{ matrix.build.os }}
 
     env:
-      RUSTFLAGS: ${{ matrix.build.rustflags }} --cfg tokio_unstable
```
i1i1, line 59: I would better compile release with debug. WDUT?

src/tests.rs
i1i1, line 1: I would better move file into `tests/` folder

ozgunozerk: that was my first attempt. Then I thought, these tests are not integration tests, but unit tests. So I followed monorepo's farmer library approach.

i1i1: Yeah, makes sense then

src/commands/farm.rs
```diff
@@ -67,7 +67,7 @@ pub(crate) async fn farm(is_verbose: bool, executor: bool) -> Result<()> {
         }
     }
 
-    let summary = Summary::new(Some(farmer_config.plot_size)).await?;
+    let summary = Summary::new(Some(farmer_config.clone().plot_size)).await?;
```
i1i1: Do you need clone here though :thinking: ?

ozgunozerk: yes, `copy` is not implemented for `bytesize` wrapper on sdk. No big deal, we can fix it later :)

i1i1: Okay, can you move clone to `plot_size` then? Like `farmer_config.plot_size.clone()`?

ozgunozerk: lemme try sure

i1i1 (dismissed):

i1i1 (approved):

PR #183, opened 2023-04-25
Add cache

ozgunozerk:
Yanking version to 0.4.0 since this is a breaking change to summary, due to a new field added.

Changes:
- summary -> this is now a cache
- summary_file -> this is the file for persistent storage for the summary info

Logic (very high level):
- on start, read the summary_file and initialize the cache
- on events, update the cache accordingly
- at every 100 block, write cache to persistent storage

### DIFF
--- a/src/commands/farm.rs
+++ b/src/commands/farm.rs
@@ -14,7 +14,7 @@ use tokio::task::JoinHandle;
 use tracing::instrument;
 
 use crate::config::{validate_config, ChainConfig, Config};
-use crate::summary::{Rewards, Summary, SummaryFilePointer, SummaryUpdateFields};
+use crate::summary::{Rewards, Summary, SummaryFile, SummaryInner, SummaryUpdateFields};
 use crate::utils::{install_tracing, raise_fd_limit, spawn_task, IntoEyre, IntoEyreStream};
 
 /// allows us to detect multiple instances of the farmer and act on it
@@ -67,7 +67,7 @@ pub(crate) async fn farm(is_verbose: bool, executor: bool) -> Result<()> {
         }
     }
 
-    let summary = SummaryFilePointer::new(Some(farmer_config.plot_size)).await?;
+    let summary = Summary::new(SummaryFile::new(Some(farmer_config.plot_size)).await?).await?;
 
     println!("Starting farmer ...");
     let farmer = Arc::new(farmer_config.build(&node).await?);
@@ -194,7 +194,7 @@ async fn subscribe_to_node_syncing(node: &Node) -> Result<()> {
 }
 
 async fn subscribe_to_plotting_progress(
-    summary: SummaryFilePointer,
+    mut summary: Summary,
     farmer: Arc<Farmer>,
     is_initial_progress_finished: Arc<AtomicBool>,
     sector_size_bytes: u64,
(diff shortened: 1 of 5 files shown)


src/commands/farm.rs
```diff
@@ -239,30 +239,28 @@ async fn subscribe_to_plotting_progress(
...
+            .get_events(Some(block_hash))
             .await
             .into_eyre()
-            .context("couldn't get events from node")?;
-
-        for event in events {
-            // subscription is active when plotting is started, only print out rewards after
-            // plotting finishes to not corrup the progress bars
+            .context("couldn't get event from node")?
+        {
```
i1i1, line 263: I would parallelize it using [async_stream](https://docs.rs/async-stream/latest/async_stream/):
```rust
    fn missed_blocks_stream(
        node: std::sync::Arc<Node>,
        mut last_block_num: subspace_sdk::node::BlockNumber,
    ) -> impl Stream<Item = anyhow::Result<subspace_sdk::node::BlockNumber>> {
        async_stream::try_stream! {
            loop {
                let last_retrieved_num = node.get_info().await?.finalized_block.1;

                if last_block_num == last_retrieved_num {
                    break;
                }

                while last_block_num < last_retrieved_num {
                    yield last_block_num;
                    last_block_num += 1;
                }
            }
        }
    }

    let concurency_parameter = todo!();

    let node = std::sync::Arc::new(todo!());
    let s = missed_blocks_stream(std::sync::Arc::clone(&node), todo!());

    futures::pin_mut!(s);

    let missed_rewards: u128 = s
        .map(|block_number| {
            Ok(node.get_events(Some(
                node.block_hash(block_number?)?
                    .expect("Should be always present if no pruning enabled"),
            )))
        })
        .try_buffer_unordered(concurency_parameter)
        .map_ok(|events| futures::stream::iter(events).map(anyhow::Result::Ok))
        .try_flatten()
        .map_ok(|event| match event {
            Event::Rewards(
                RewardsEvent::VoteReward { voter: author, reward }
                | RewardsEvent::BlockReward { block_author: author, reward },
            ) if author == reward_address.into() => reward,
            _ => 0,
        })
        .try_fold(0u128, |acc, rewards| {
            futures::future::ready::<anyhow::Result<_>>(Ok(acc + rewards))
        })
        .await?;
```

i1i1: Sorry, missed part with farmed block count here, but I think you get an idea.

src/summary.rs
```diff
@@ -40,39 +40,114 @@ impl std::fmt::Display for Rewards {
...
-/// command
-#[derive(Deserialize, Serialize, Debug)]
+/// struct for flexibly updating the fields of the summary
+#[derive(Default, Debug)]
+pub(crate) struct SummaryUpdateFields {
+    pub(crate) is_plotting_finished: bool,
+    pub(crate) is_new_block_farmed: bool,
+    pub(crate) is_new_vote: bool,
+    pub(crate) maybe_new_reward: Option<Rewards>,
+    pub(crate) maybe_last_block: Option<u64>,
```
i1i1, line 50: I think you should cache these 4 things:
- last seen finalized block
- total votes
- total farmed blocks
- total rewards

src/commands/farm.rs
i1i1: You can update summary somewhere in between here

src/summary.rs
ozgunozerk: but I do require `is_plotting_finished` as well

i1i1, line 50:
```suggestion
    pub(crate) is_plotting_finished: bool,
    pub(crate) new_block_farmed: u32,
    pub(crate) new_vote: bool,
    pub(crate) new_rewards: Rewards,
    pub(crate) maybe_last_block: Option<u64>,
```

src/summary.rs
```diff
@@ -40,39 +40,114 @@ impl std::fmt::Display for Rewards {
...
+impl SummaryFile {
     /// creates new summary file pointer
     ///
     /// if user_space_pledged is provided, it creates a new summary file
     /// else, it tries to read the existing summary file
     #[instrument]
-    pub(crate) async fn new(user_space_pledged: Option<ByteSize>) -> Result<SummaryFilePointer> {
+    pub(crate) async fn new(user_space_pledged: Option<ByteSize>) -> Result<SummaryFile> {
         let summary_path = summary_path();
         let summary_dir = dirs::data_local_dir()
```
i1i1, line 152: Use `dirs::cache_dir` here

src/summary.rs
ozgunozerk: we already have this struct as `Summary`
this struct was intermediate for updating the summary.
After our discussion, it probably won't be used

ozgunozerk:
Reworking, will re-open when I'm done, don't wanna notify with push messages

PR #184, opened 2023-04-27
Add cache

ozgunozerk:
This PR:
- Adds cache mechanism. Which updates the summary file in every 100 blocks scanned
- instead of relying on only new events, now we are scanning the whole blockchain from start to tip, and saving our progress to summary file at every 100 block
- this process of scanning the blockchain is parallelized (thanks a lot @i1i1 )
- revamps the summary logic w.r.t above changes
- updates SDK revision

It should be more reliable and safe compared to previous approach we were following (taking into account only new events, and thus, being more local, and prone to edge scenarios and bugs)

Will squash merge, so review all at once if you like

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,6 +35,7 @@ tracing-subscriber = { version = "0.3.16", features = ["env-filter"] }
 whoami = "1"
 
 subspace-sdk = { git = "https://github.com/subspace/subspace-sdk", rev = "ae2c1f6bd975f3471ee0d7243f2753b10593e194" }
+derive_more = "0.99.17"
 
 # The only triple tested and confirmed as working in `jemallocator` crate is `x86_64-unknown-linux-gnu`
 [target.'cfg(all(target_arch = "x86_64", target_vendor = "unknown", target_os = "linux", target_env = "gnu"))'.dependencies]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,7 +36,6 @@ tracing-subscriber = { version = "0.3.16", features = ["env-filter"] }
 whoami = "1"
 
 subspace-sdk = { git = "https://github.com/subspace/subspace-sdk", rev = "ae2c1f6bd975f3471ee0d7243f2753b10593e194" }
-sp-core = { version = "7.0.0", git = "https://github.com/subspace/substrate", rev = "fdb68194ab6995447610b3dbdee70559711dbd63" }
 
 # The only triple tested and confirmed as working in `jemallocator` crate is `x86_64-unknown-linux-gnu`
 [target.'cfg(all(target_arch = "x86_64", target_vendor = "unknown", target_os = "linux", target_env = "gnu"))'.dependencies]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,7 +31,7 @@ tracing-error = "0.2.0"
 tracing-subscriber = { version = "0.3.16", features = ["env-filter"] }
 whoami = "1"
 
-subspace-sdk = { git = "https://github.com/subspace/subspace-sdk", rev = "3bccdb82af5dd5de7d84867cbb3c7ad5b316a13e" }
+subspace-sdk = { git = "https://github.com/subspace/subspace-sdk", rev = "ac22c753a3e6f0a4fbbc4255da1678297ed93fa3" }
 termion = "2.0.1"
 open = "4.0.2"
 
@@ -127,7 +127,10 @@ zeroize = { opt-level = 3 }
 frame-benchmarking = { version = "4.0.0-dev", git = "https://github.com/subspace/substrate", rev = "fdb68194ab6995447610b3dbdee70559711dbd63" }
 frame-support = { version = "4.0.0-dev", git = "https://github.com/subspace/substrate", rev = "fdb68194ab6995447610b3dbdee70559711dbd63" }
 frame-system = { version = "4.0.0-dev", git = "https://github.com/subspace/substrate", rev = "fdb68194ab6995447610b3dbdee70559711dbd63" }
+sp-api = { version = "4.0.0-dev", git = "https://github.com/subspace/substrate", rev = "fdb68194ab6995447610b3dbdee70559711dbd63" }
 sp-core = { version = "7.0.0", git = "https://github.com/subspace/substrate", rev = "fdb68194ab6995447610b3dbdee70559711dbd63" }
+sp-inherents = { version = "4.0.0-dev", git = "https://github.com/subspace/substrate", rev = "fdb68194ab6995447610b3dbdee70559711dbd63" }
+sp-state-machine = { version = "0.13.0", git = "https://github.com/subspace/substrate", rev = "fdb68194ab6995447610b3dbdee70559711dbd63" }
 sp-std = { version = "5.0.0", git = "https://github.com/subspace/substrate", rev = "fdb68194ab6995447610b3dbdee70559711dbd63" }
 sp-io = { version = "7.0.0", git = "https://github.com/subspace/substrate", rev = "fdb68194ab6995447610b3dbdee70559711dbd63" }
 sp-runtime = { version = "7.0.0", git = "https://github.com/subspace/substrate", rev = "fdb68194ab6995447610b3dbdee70559711dbd63" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,6 +5,7 @@ edition = "2021"
 
 [dependencies]
 anyhow = "1"
+async-stream = "0.3.5"
 bytesize = "1.1"
 bytesize-serde = "0.2"
 clap = { version = "4.1.1", features = ["derive"] }
@@ -15,12 +16,14 @@ fdlimit = "0.2"
 futures = "0.3"
 indicatif = { version = "0.17.1", features = ["improved_unicode"] }
 libp2p-core = "0.38"
+open = "4.0.2"
 owo-colors = "3.5.0"
 serde = "1"
 serde_derive = "1"
 single-instance = "0.3.3"
 strum = "0.24.1"
 strum_macros = "0.24.3"
+termion = "2.0.1"
 thiserror = "1"
 toml = "0.7"
 tokio = { version = "1.27", features = ["macros", "rt-multi-thread", "tracing"] }
@@ -32,8 +35,7 @@ tracing-subscriber = { version = "0.3.16", features = ["env-filter"] }
 whoami = "1"
 
 subspace-sdk = { git = "https://github.com/subspace/subspace-sdk", rev = "ac22c753a3e6f0a4fbbc4255da1678297ed93fa3" }
-termion = "2.0.1"
-open = "4.0.2"
+sp-consensus-subspace = { git = "https://github.com/subspace/subspace", rev = "965813c19810f8402a6022c2ffd70e66cb269759" }
 
 # The only triple tested and confirmed as working in `jemallocator` crate is `x86_64-unknown-linux-gnu`
 [target.'cfg(all(target_arch = "x86_64", target_vendor = "unknown", target_os = "linux", target_env = "gnu"))'.dependencies]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,8 +34,7 @@ tracing-error = "0.2.0"
 tracing-subscriber = { version = "0.3.16", features = ["env-filter"] }
 whoami = "1"
 
-subspace-sdk = { git = "https://github.com/subspace/subspace-sdk", rev = "ac22c753a3e6f0a4fbbc4255da1678297ed93fa3" }
-tokio-stream = "0.1.13"
+subspace-sdk = { git = "https://github.com/subspace/subspace-sdk", rev = "ae2c1f6bd975f3471ee0d7243f2753b10593e194" }
 
 # The only triple tested and confirmed as working in `jemallocator` crate is `x86_64-unknown-linux-gnu`
 [target.'cfg(all(target_arch = "x86_64", target_vendor = "unknown", target_os = "linux", target_env = "gnu"))'.dependencies]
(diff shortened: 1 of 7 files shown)


src/summary.rs
```diff
@@ -40,41 +41,57 @@ impl std::fmt::Display for Rewards {
...
+    }
 }
 
 /// utilizing persistent storage for the information to be displayed for the
 /// `info` command
 #[derive(Debug, Clone)]
-pub(crate) struct SummaryFilePointer {
-    file: Arc<Mutex<PathBuf>>,
+pub(crate) struct SummaryFile {
+    inner: Arc<Mutex<PathBuf>>,
```
i1i1, line 83: I think you either need to have `Arc<Mutex<File>>` here or just a `PathBuf`. `PathBuf` is not modified at all. I would prefer having wrapped `File` here

i1i1 (changes requested):
Otherwise, looking good!

src/commands/farm.rs
```diff
@@ -347,3 +412,23 @@ fn syncing_progress_bar(current_block: u64, total_blocks: u64) -> ProgressBar {
     pb.set_position(current_block);
     pb
 }
+
+fn not_yet_processed_block_nums_stream(
```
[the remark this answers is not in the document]
ozgunozerk: these are actually not new blocks. Syncing maybe ahead of this. The name is long, I agree, but it's clear. 
`new_blocks` are confusing imo. 2 months later we can forget the distinction and it may not make sense

src/summary.rs
```diff
@@ -40,41 +41,57 @@ impl std::fmt::Display for Rewards {
...
-pub(crate) struct SummaryUpdateFields {
-    pub(crate) is_plotting_finished: bool,
-    pub(crate) is_new_block_farmed: bool,
-    pub(crate) is_new_vote: bool,
-    pub(crate) maybe_new_reward: Option<Rewards>,
+impl Summary {
+    #[instrument]
+    pub(crate) async fn new(summary_file: SummaryFile) -> Result<Summary> {
+        summary_file.parse().await
+    }
```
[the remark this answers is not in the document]
ozgunozerk: true, that was an artifact.

src/summary.rs
ozgunozerk: The reason for `Arc<Mutex<PathBuf>>` was to measure against data races. There are different threads/tasks trying to reach this. 

I see your point, `PathBuf` itself is not being modified. However, by keeping it in a `Mutex`, we are giving the `key` to summary to only one thread/task. 

So, `PathBuf` alone (without Mutex) would not be safe. 
And I didn't want to put `File` in there, since I don't want to keep file open.

src/commands/farm.rs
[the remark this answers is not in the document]
i1i1: Okay, so be it!

src/summary.rs
i1i1: > since I don't want to keep file open

Why so? I think it is should be fine

src/summary.rs
```diff
@@ -40,41 +41,57 @@ impl std::fmt::Display for Rewards {
...
+}
+
+/// struct for flexibly updating the fields of the summary
+#[derive(Default, Debug)]
+pub(crate) struct SummaryUpdateFields {
+    pub(crate) is_plotting_finished: bool,
+    pub(crate) maybe_authored_count: Option<u64>,
+    pub(crate) maybe_vote_count: Option<u64>,
+    pub(crate) maybe_reward: Option<Rewards>,
+    pub(crate) maybe_new_blocks: Option<BlockNumber>,
```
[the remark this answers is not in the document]
i1i1: > And supplying update(0, 0, 0, Rewards(0), true), seems weird in terms of API, compared to update(None, None, None, None, true)

I think it is fine. If 0 blocks were authored, it means that no blocks were authored. Actually `Some(0)` would look weird. So if 0 blocks were authored, you don't need to update a field. Same goes for other fields (apart from `last_block`)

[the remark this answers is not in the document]
ozgunozerk: There won't be any `Some(0)` case?

src/summary.rs
ozgunozerk: because we have `info` command which can be run simultenaously whilst running a farmer. And `info` command reads this summary file

src/commands/farm.rs
```diff
@@ -283,14 +314,16 @@ async fn subscribe_to_solutions(
                     // - Number of times we authored a block
                     .map(|hash| {
                         let is_author = node_clone
-                            .block_header(hash)?
+                            .block_header(hash)
+                            .into_eyre()?
```
i1i1: `.context()` is missing here

src/commands/farm.rs
```diff
@@ -261,14 +277,29 @@ async fn subscribe_to_solutions(
...
+            }
         })
+        .filter_map(|res| async {
+            match res {
+                Ok(Some(block_hash)) => Some(block_hash),
+                Ok(None) => None,
+                Err(err) => Some(future::ready(Err(err))),
+            }
+        })
+        .map(|x| x.into_inner())
```
i1i1: Sorry, I've missed it before. I think you might want this, here:
```rust
.try_filter_map(|block| match node.block_hash(block).into_eyre() {
    Ok(Some(block_hash)) => future::ok(Some(block_hash)),
    Ok(None) if blocks_pruning => future::ok(None),
    Ok(None) => future::err(eyre!("node database is probably corrupted, try wiping the node")),
    Err(err) => future::err(err.context("add context here")),
})
```

src/commands/farm.rs
```diff
@@ -261,14 +277,29 @@ async fn subscribe_to_solutions(
...
         // Chunk block hashes in chunks of `n_blocks`
         .try_chunks(n_blocks)
-        .map_err(anyhow::Error::from)
+        .map_err(Error::from)
```
i1i1: I would also add context here

src/commands/farm.rs
```diff
@@ -335,20 +368,18 @@ async fn subscribe_to_solutions(
```
i1i1: And here

src/commands/farm.rs
```diff
@@ -415,10 +446,10 @@ fn syncing_progress_bar(current_block: u64, total_blocks: u64) -> ProgressBar {
 fn not_yet_processed_block_nums_stream(
     node: std::sync::Arc<Node>,
     mut last_block_num: subspace_sdk::node::BlockNumber,
-) -> impl Stream<Item = anyhow::Result<subspace_sdk::node::BlockNumber>> {
+) -> impl Stream<Item = Result<subspace_sdk::node::BlockNumber>> {
     async_stream::try_stream! {
         loop {
-            let last_retrieved_num = node.get_info().await?.finalized_block.1;
+            let last_retrieved_num = node.get_info().await.into_eyre()?.finalized_block.1;
```
i1i1, line 365: And here :(

src/summary.rs
```diff
@@ -17,44 +18,17 @@ use tokio::sync::Mutex;
...
 pub(crate) struct SummaryUpdateFields {
     pub(crate) is_plotting_finished: bool,
-    pub(crate) maybe_authored_count: Option<u64>,
-    pub(crate) maybe_vote_count: Option<u64>,
-    pub(crate) maybe_reward: Option<Rewards>,
-    pub(crate) maybe_new_blocks: Option<BlockNumber>,
+    pub(crate) new_authored_count: u64,
+    pub(crate) new_vote_count: u64,
+    pub(crate) new_reward: Rewards,
+    pub(crate) new_parsed_blocks: BlockNumber,
```
i1i1, line 31: `new_` is a bit ambiguous here. It is either a new value which we need to set, or value which we need to add. I would just remove it.

src/commands/farm.rs
```diff
@@ -265,138 +268,48 @@ async fn subscribe_to_solutions(
...
+    // after that, there will be new blocks arrived
+    // process these new blocks sequentially
+    process_block_stream(
+        last_block_num,
+        node.clone(),
+        blocks_pruning,
+        summary_file.clone(),
+        reward_address,
+        1000,
+        10,
```
i1i1, line 281: I would extract those to constants

src/commands/farm.rs
```diff
@@ -462,3 +375,130 @@ fn not_yet_processed_block_nums_stream(
...
+    let (rewards, votes, author) = futures::stream::iter(blocks)
+        // We scan each hash and find 3 things:
+        // - Total amount of rewards
+        // - Number of votes
+        // - Number of times we authored a block
+        .map(|hash| {
+            let is_author = node
+                .block_header(hash)
+                .into_eyre()?
+                .expect("TODO: Account for missing blocks somehow or check pruning")
```
i1i1, line 465: You need to check pruning here :D

src/commands/farm.rs
```diff
@@ -462,3 +375,130 @@ fn not_yet_processed_block_nums_stream(
...
+) -> Result<(u128, u64, u64)> {
+    let (rewards, votes, author) = futures::stream::iter(blocks)
+        // We scan each hash and find 3 things:
+        // - Total amount of rewards
+        // - Number of votes
+        // - Number of times we authored a block
+        .map(|hash| {
+            let is_author = node
+                .block_header(hash)
+                .into_eyre()?
```
i1i1, line 464: context here

src/commands/farm.rs
```diff
@@ -462,3 +375,130 @@ fn not_yet_processed_block_nums_stream(
...
+
+            Result::Ok(rewards_future)
+        })
+        // We calculate each block in parallel
+        .try_buffer_unordered(n_tasks)
+        // After that we sum up result
+        .try_fold((0, 0, 0), |(rewards, votes, author), (new_rewards, new_votes, new_author)| {
+            futures::future::ok((rewards + new_rewards, votes + new_votes, author + new_author))
+        })
+        .await?;
```
i1i1, line 501: and here

src/summary.rs
ozgunozerk: > new_ is a bit ambiguous here. It is either a new value which we need to set, or value which we need to add. I would just remove it.

`new_authored_count` is not ambiguous imo. In fact, removing `new` would make it ambiguous like:
>  It is either a new value which we need to set, or value which we need to add.

`new` removes the ambiguity and signals that we have to add them, since they are new.

i1i1: You right here. Closing this thread

src/commands/farm.rs
```diff
@@ -35,7 +38,7 @@ type MaybeHandles = Option<(JoinHandle<Result<()>>, JoinHandle<Result<()>>)>;
 #[instrument]
 pub(crate) async fn farm(is_verbose: bool, executor: bool) -> Result<()> {
     install_tracing(is_verbose);
-    color_eyre::install()?;
+    color_eyre::install().expect("first installment always succeeds");
```
i1i1, line 41: I would prefer not to panic here :(

src/commands/farm.rs
```diff
@@ -390,31 +398,19 @@ async fn process_block_stream(
...
-                        future::ok(None)
-                    },
-            }
+        .try_filter_map(|block| match node.block_hash(block).into_eyre() {
+            Ok(Some(block_hash)) => future::ok(Some(block_hash)),
+            Ok(None) if blocks_pruning => future::ok(None),
+            Ok(None) =>
+                future::err(eyre!("node database is probably corrupted, try wiping the node")),
+            // Err(err) => future::err(err.context("block hash couldn't found")),
+            Err(err) => future::err(eyre!("couldn't get block hash from node, because: {err}")),
```
i1i1, line 409: I think `Err(err) => future::err(err.wrap_err("block hash couldn't found"))` would work here.

It is really important to keep context everywhere. `eyre!` macro here will just expand latest error message, discarding any other ones

src/commands/farm.rs
```diff
@@ -390,31 +398,19 @@ async fn process_block_stream(
...
-                Ok(Some(block_hash)) => Some(block_hash),
-                Ok(None) => None,
-                Err(err) => Some(future::ready(Err(err))),
-            }
-        })
-        .map(|x| x.into_inner())
         // Chunk block hashes in chunks of `n_blocks`
         .try_chunks(batch_blocks)
+        .map(|x| x.context("try chunks failed"))
         .map_err(Error::from)
```
i1i1: Seems a bit redundant, right now. I think you can use:
```rust
.map_err(|err| err.wrap_err("Fetching blocks failed"))
```

i1i1 (commented):
Otherwise looking good!

src/commands/farm.rs
```diff
@@ -2,23 +2,29 @@ use std::io::Write;
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Arc;
 
-use color_eyre::eyre::{eyre, Context, Result};
+use color_eyre::eyre::{eyre, Context, Error, Result};
 use futures::prelude::*;
 use indicatif::{ProgressBar, ProgressStyle};
 use owo_colors::OwoColorize;
 use single_instance::SingleInstance;
+use sp_core::H256;
```
i1i1, line 10: I think you don't need it. You should use `subspace_sdk::node::Hash` type alias.

Cargo.toml
```diff
@@ -31,9 +35,8 @@ tracing-error = "0.2.0"
 tracing-subscriber = { version = "0.3.16", features = ["env-filter"] }
 whoami = "1"
 
-subspace-sdk = { git = "https://github.com/subspace/subspace-sdk", rev = "3bccdb82af5dd5de7d84867cbb3c7ad5b316a13e" }
-termion = "2.0.1"
-open = "4.0.2"
+subspace-sdk = { git = "https://github.com/subspace/subspace-sdk", rev = "ae2c1f6bd975f3471ee0d7243f2753b10593e194" }
+sp-core = { version = "7.0.0", git = "https://github.com/subspace/substrate", rev = "fdb68194ab6995447610b3dbdee70559711dbd63" }
```
i1i1, line 39: You should remove this one

src/commands/farm.rs
```diff
@@ -347,3 +364,148 @@ fn syncing_progress_bar(current_block: u64, total_blocks: u64) -> ProgressBar {
...
+
+    let mut current_iter = 0;
+
+    stream
+        .try_filter_map(|block| match node.block_hash(block).into_eyre() {
+            Ok(Some(block_hash)) => future::ok(Some(block_hash)),
+            Ok(None) if blocks_pruning => future::ok(None),
+            Ok(None) =>
+                future::err(eyre!("node database is probably corrupted, try wiping the node")),
+            // Err(err) => future::err(err.context("block hash couldn't found")),
```
i1i1, line 409: You can remove this comment right now

i1i1 (approved):

i1i1 (approved):

PR #200, opened 2023-05-15
key generation if user does not have a public key beforehand

### LINKED ISSUE #199: 🎁 [Feature Request]: Generate mnemonic and public key in CLI if user wants to

1. CLI should ask the the user if they have an existing public key
2. if not, a new prompt for if user wants to generate a public key
3. display the mnemonic, then redact it, and prompt user to prove if they stored the mnemonic (ask random words from the 12 word mnemonic)
4. display the public key, and set it

### Code of Conduct

ozgunozerk:
This PR adds a new feature, which is having the ability to generate public key and mnemonic in the CLI.
Closes #199 

New flow for the farmer address during `init` step:

1. CLI asks the the user if they have an existing public key
2. if not, a new prompt for if user wants to generate a public key
3. display the mnemonic, show the proper messages, then redact it, and prompt user to prove if they stored the mnemonic (ask random words from the 12 word mnemonic)
4. if users passes the test, display the public key, and set it

Redacting the mnemonic was important for this setup, and I'm using crossterm to handle edge cases here. Code should be safe for all sized terminals for redacting the mnemonic completely without clearing other parts of the terminal.

### DIFF
--- a/src/commands/init.rs
+++ b/src/commands/init.rs
@@ -1,8 +1,13 @@
-use std::io::Write;
+use std::io::{BufRead, Write};
 use std::str::FromStr;
 
-use color_eyre::eyre::{Context, Result};
+use color_eyre::eyre::{eyre, Context, Error, Result};
+use crossterm::terminal::{Clear, ClearType};
+use crossterm::{cursor, execute};
+use rand::prelude::IteratorRandom;
+use sp_core::Pair;
 use strum::IntoEnumIterator;
+use subspace_sdk::PublicKey;
 
 use crate::config::{
     create_config, AdvancedFarmerSettings, AdvancedNodeSettings, ChainConfig, Config, FarmerConfig,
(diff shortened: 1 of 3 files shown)


src/commands/init.rs
```diff
@@ -126,73 +127,70 @@ fn get_config_from_user_inputs() -> Result<Config> {
...
-            // print the public key and return it
-            println!("Your new public key is: {}", pair.public());
-            let public_key_array = pair.public().0;
-            Ok(public_key_array.into())
-        } else {
-            Err(eyre!("New key creation was not confirmed"))
+    // generate new mnemonic and key pair
+    let (pair, mut phrase, _): (sp_core::sr25519::Pair, String, _) =
+        Pair::generate_with_phrase(None);
+    let words: Vec<&str> = phrase.split_whitespace().collect();
```
i1i1, line 146: I would just wrap every data structure in `zeroize::Zeroizing`, so that during drop it will zeroize everything

PR #212, opened 2023-06-20
create actions workflow self-hosted runner

DaMandal0rian:
This PR creates self-hosted runners for the subspace cli repository that will run jobs on the following:
- Linux X86_64
- Mac ARM64
- Windows x86_64

Cross compilation of AARCH64 linux will be done on x86_64.

### DIFF
--- a/.github/workflows/release-self-hosted.yml
+++ b/.github/workflows/release-self-hosted.yml
@@ -1,0 +1,213 @@
+# This action enables building the executables for cli, can be triggered manually or by release creation.
+name: release
+
+on:
+  workflow_dispatch:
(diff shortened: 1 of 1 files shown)


ozgunozerk:
I couldn't see an action with this, is this tested? And master needs to be replaced with main in the new workflow

DaMandal0rian:
> I couldn't see an action with this, is this tested? And master needs to be replaced with main in the new workflow

Yes I had a comment to replace master with main after testing, but since it will be merged into develop branch first, I updated it.

ozgunozerk (approved):

PR #225, opened 2023-07-07
CI update to use GH self-hosted runners

DaMandal0rian:
This PR introduces the use of self hosted runners for the repository.
- supports Ubuntu-22.04, macOS-12 and Windows-server-2022 runners

### DIFF
--- a/.github/workflows/ci-tests.yml
+++ b/.github/workflows/ci-tests.yml
@@ -32,10 +32,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os:
-          - ubuntu-22.04
-          - macos-12
-          - windows-2022
+        os: ${{ fromJson(github.repository_owner == 'subspace' && github.repository == 'subspace/subspace-cli' && '[["self-hosted", "ubuntu-22.04-x86-64", "Linux"], ["self-hosted", "macos-12-arm64", "macOS"], ["self-hosted", "windows-server-2022-x86-64", "Windows"]]' || '["ubuntu-22.04", "macos-12", "windows-2022"]') }}
         run-all:
           - ${{ inputs.test-macos-and-windows == true || github.ref == 'refs/heads/main' }}
         exclude: # exclude macos-12 and windows-2022 when the condition is false
@@ -43,28 +40,40 @@ jobs:
             os: macos-12
           - run-all: false
             os: windows-2022
+          - run-all: false
+            os: macOS
+          - run-all: false
+            os: Windows
 
     runs-on: ${{ matrix.os }}
     steps:
       - name: git checkout
         uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8 # @v3.1.0
 
-      # On macOS, we need a proper Clang version, not Apple's custom version without wasm32 support
+      # On macOS, we need a proper Clang version, not Apple's custom version without wasm32 support, we also install clang on Windows
       - name: Install LLVM and Clang
         uses: KyleMayes/install-llvm-action@8852e4d5c58653ed05135c0a5d949d9c2febcb00 # v1.6.1
         with:
           version: "15.0"
-        if: runner.os == 'macOS'
+        if: runner.os == 'macOS' || runner.os == 'Windows'
 
       - name: Install Protoc
-        uses: arduino/setup-protoc@v1
+        uses: arduino/setup-protoc@v2
         with:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
+        if: runner.os == 'Linux' || runner.os == 'macOS'
 
-      # Workaround to resolve link error with C:\msys64\mingw64\bin\libclang.dll
-      - name: Remove msys64
-        run: Remove-Item -LiteralPath "C:\msys64\" -Force -Recurse
+      - name: Install Protoc Windows
+        run: |
+          Invoke-WebRequest -Uri https://github.com/protocolbuffers/protobuf/releases/download/v23.3/protoc-23.3-win64.zip -Outfile protoc-23.3-win64.zip
+          Expand-Archive -Force .\protoc-23.3-win64.zip -DestinationPath C:\protoc-23.3 
         if: runner.os == 'Windows'
+
+      - name: Install nightly
+        uses: actions-rs/toolchain@v1
+        with:
+            toolchain: nightly-2023-05-16
+            components: rustfmt, clippy, rust-src,rust-std
 
       - name: cargo fmt
         uses: actions-rs/cargo@ae10961054e4aa8b4aa7dffede299aaf087aa33b # @v1.0.1
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -17,37 +17,37 @@ jobs:
     strategy:
       matrix:
         build:
-          - os: ubuntu-20.04
+          - os: ${{ fromJson(github.repository_owner == 'subspace' && github.repository == 'subspace/subspace-cli' && '["self-hosted", "ubuntu-22.04-x86-64", "Linux"]' || '["ubuntu-22.04"]') }}
             target: x86_64-unknown-linux-gnu
             production_target: target/x86_64-unknown-linux-gnu/production
             suffix: ubuntu-x86_64-v2-${{ github.ref_name }}
             rustflags: "-C target-cpu=x86-64-v2"
-          - os: ubuntu-20.04
+          - os: ${{ fromJson(github.repository_owner == 'subspace' && github.repository == 'subspace/subspace-cli' && '["self-hosted", "ubuntu-22.04-x86-64", "Linux"]' || '["ubuntu-22.04"]') }}
             target: x86_64-unknown-linux-gnu
             production_target: target/x86_64-unknown-linux-gnu/production
             suffix: ubuntu-x86_64-skylake-${{ github.ref_name }}
             rustflags: "-C target-cpu=skylake"
-          - os: ubuntu-20.04
+          - os: ${{ fromJson(github.repository_owner == 'subspace' && github.repository == 'subspace/subspace-cli' && '["self-hosted", "ubuntu-22.04-x86-64", "Linux"]' || '["ubuntu-22.04"]') }}
             target: aarch64-unknown-linux-gnu
             production_target: target/aarch64-unknown-linux-gnu/aarch64linux
             suffix: ubuntu-aarch64-${{ github.ref_name }}
             rustflags: "-C linker=aarch64-linux-gnu-gcc"
-          - os: macos-12
+          - os: ${{ fromJson(github.repository_owner == 'subspace' && github.repository == 'subspace/subspace-cli' && '["self-hosted", "macos-12-arm64", "macOS"]' || '["macos-12"]') }}
             target: x86_64-apple-darwin
             production_target: target/x86_64-apple-darwin/production
             suffix: macos-x86_64-${{ github.ref_name }}
             rustflags: ""
-          - os: macos-12
+          - os: ${{ fromJson(github.repository_owner == 'subspace' && github.repository == 'subspace/subspace-cli' && '["self-hosted", "macos-12-arm64", "macOS"]' || '["macos-12"]') }}
             target: aarch64-apple-darwin
             production_target: target/aarch64-apple-darwin/production
             suffix: macos-aarch64-${{ github.ref_name }}
             rustflags: ""
-          - os: windows-2022
+          - os: ${{ fromJson(github.repository_owner == 'subspace' && github.repository == 'subspace/subspace-cli' && '["self-hosted", "windows-server-2022-x86-64", "Windows"]' || '["windows-2022"]') }}
             target: x86_64-pc-windows-msvc
             production_target: target/x86_64-pc-windows-msvc/production
             suffix: windows-x86_64-v2-${{ github.ref_name }}
             rustflags: "-C target-cpu=x86-64-v2"
-          - os: windows-2022
+          - os: ${{ fromJson(github.repository_owner == 'subspace' && github.repository == 'subspace/subspace-cli' && '["self-hosted", "windows-server-2022-x86-64", "Windows"]' || '["windows-2022"]') }}
             target: x86_64-pc-windows-msvc
             production_target: target/x86_64-pc-windows-msvc/production
             suffix: windows-x86_64-skylake-${{ github.ref_name }}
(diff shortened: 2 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -17,37 +17,37 @@ jobs:
     strategy:
       matrix:
         build:
-          - os: ubuntu-20.04
+          - os: ${{ fromJson(github.repository_owner == 'subspace' && github.repository == 'subspace/subspace-cli' && '["self-hosted", "ubuntu-22.04-x86-64", "Linux"]' || '["ubuntu-22.04"]') }}
```
jfrank-summit, line 20: Is there any way to abstract this logic out so that it doesn't have to be repeated?

jfrank-summit (commented):
Looks good. I had one comment/question. Also, there is a problem with `clippy-fmt-test (ubuntu-22.04, false) Expected — Waiting for status to be reported`.

DaMandal0rian:
> Looks good. I had one comment/question. Also, did you test this on a fork of the repo?

Yes, was not able to run it on a fork

> Looks good. I had one comment/question. Also, did you test this on a fork of the repo?

I tested on a fork and was not able to execute the runner due to `github.repository_owner == 'subspace' && github.repository == 'subspace/subspace-cli'`

So it will fall back to using github hosted actions.

.github/workflows/release.yml
DaMandal0rian: Unfortunately, GitHub doesn't support yaml anchors still :(
see https://github.com/orgs/community/discussions/25545
and https://github.com/actions/runner/issues/1182

jfrank-summit (approved):
lgtm

.github/workflows/ci-tests.yml
```diff
@@ -32,40 +32,49 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os:
-          - ubuntu-22.04
-          - macos-12
-          - windows-2022
+        os: ${{ fromJson(github.repository_owner == 'subspace' && github.repository == 'subspace/subspace-cli' && '[["self-hosted", "ubuntu-22.04-x86-64", "Linux"], ["self-hosted", "macos-12-arm64", "macOS"], ["self-hosted", "windows-server-2022-x86-64", "Windows"]]' || '["ubuntu-22.04", "macos-12", "windows-2022"]') }}
```
nazar-pc, line 35: `&& github.repository == 'subspace/subspace-cli'` is unnecessary

nazar-pc: Using `"self-hosted", "ubuntu-22.04-x86-64", "Linux"` is unnecessary as well, it is very obvious that Ubuntu is Linux, so it should have been `"self-hosted", "ubuntu-22.04-x86-64"`

.github/workflows/ci-tests.yml
```diff
@@ -32,40 +32,49 @@ jobs:
...
           - ${{ inputs.test-macos-and-windows == true || github.ref == 'refs/heads/main' }}
         exclude: # exclude macos-12 and windows-2022 when the condition is false
           - run-all: false
             os: macos-12
           - run-all: false
             os: windows-2022
+          - run-all: false
+            os: macOS
+          - run-all: false
+            os: Windows
```
nazar-pc, line 46: I don't think this is working, there are no such `os` values in the matrix above. You also need to find the reason these excludes are present in the first place, maybe it was because they were slow, which is not relevant anymore.

.github/workflows/ci-tests.yml
```diff
@@ -32,40 +32,49 @@ jobs:
...
 
     runs-on: ${{ matrix.os }}
     steps:
       - name: git checkout
         uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8 # @v3.1.0
 
-      # On macOS, we need a proper Clang version, not Apple's custom version without wasm32 support
+      # On macOS, we need a proper Clang version, not Apple's custom version without wasm32 support, we also install clang on Windows
```
nazar-pc, line 53: Shouldn't need to install on Windows on every run, it must be present right away just like it is with GitHub's runners.

.github/workflows/ci-tests.yml
```diff
@@ -32,40 +32,49 @@ jobs:
...
+        if: runner.os == 'Linux' || runner.os == 'macOS'
 
-      # Workaround to resolve link error with C:\msys64\mingw64\bin\libclang.dll
-      - name: Remove msys64
-        run: Remove-Item -LiteralPath "C:\msys64\" -Force -Recurse
+      - name: Install Protoc Windows
+        run: |
+          Invoke-WebRequest -Uri https://github.com/protocolbuffers/protobuf/releases/download/v23.3/protoc-23.3-win64.zip -Outfile protoc-23.3-win64.zip
+          Expand-Archive -Force .\protoc-23.3-win64.zip -DestinationPath C:\protoc-23.3 
         if: runner.os == 'Windows'
```
nazar-pc, line 70: Why two ways to install `protoc`? It should be the same way it was before. If it doesn't work then runner needs to be fixed.

nazar-pc: Also you removed "Remove msys64" step, which means you broke compatibility with GitHub's runners, while I mentioned many times it must continue to be supported like it was before.

.github/workflows/ci-tests.yml
```diff
@@ -32,40 +32,49 @@ jobs:
...
 
+      - name: Install nightly
```
nazar-pc, line 72: This step must be removed, runner must have Rust toolchain installed by default out of the box.

.github/workflows/release.yml
nazar-pc, line 20: This and below is wrong too. It must use Ubuntu 20.04, not 22.04. Using 22.04 means our executables will fail to work on Ubuntu 20.04 due to old Glibc. There is a reason behind every line and version in this file, don't change it arbitrarily.

nazar-pc (commented):
This PR broke support for GitHub runners and broke executables on Ubuntu 20.04. IMHO most of it must be reverted and our self-hosted runners should "just work" with little to no changes to workflows.

.github/workflows/ci-tests.yml
DaMandal0rian: Those are targeted from the self-hosted runner labels, which can be cross-checked with and example 
`if: runner.os == 'Windows'`

.github/workflows/ci-tests.yml
DaMandal0rian: I found a fix for this in visual studio developer tools to install llvm,clang so we don't need this anymore. Although i installed  https://visualstudio.microsoft.com/vs/features/cplusplus/ and build tools, it seems like clang is another package https://learn.microsoft.com/en-us/cpp/build/clang-support-msbuild?view=msvc-170
will test this out.

.github/workflows/ci-tests.yml
DaMandal0rian: It wasn't picking up cargo on system so as a workaround I used nightly temporarily as mentioned before, albeit it's not sustainable. I think i may have found the fix for this and will test it.

.github/workflows/release.yml
DaMandal0rian: :cry:  will setup a new linux runner with the older version, didn't know the 20.04 is a hard requirement.

.github/workflows/ci-tests.yml
DaMandal0rian: Ok will remove `&& github.repository == 'subspace/subspace-cli'` 

The github runners are funny with labels, and although it's obvious ubuntu is Linux, using "Linux" for conditional statements is cleaner.  i.e `if: runner.os == 'Linux'`

DaMandal0rian (commented):
Will look into these changes

.github/workflows/ci-tests.yml
DaMandal0rian: The `msys64` package part was causing the windows jobs to fail since the directory didn't exist. The package is used to install clang, I will add it back and include an option to only execute the step when runner is not self-hosted.

DaMandal0rian: For `protoc` the actions was failing on windows and I reported the bug upstream, so I used another method for Windows. However, I've installed `protoc`  on the windows machine and added it to the path, will test it and remove the step.

.github/workflows/ci-tests.yml
nazar-pc: You may be confusing `matrix.os` vs `runner.os`, which are not the same thing. Here we deal with `matrix.os`.

.github/workflows/ci-tests.yml
nazar-pc: I'm not sure what MSVC has to do with LLVM, I'd expect you to install it from here: https://github.com/llvm/llvm-project/releases/tag/llvmorg-15.0.7 (latest release of 15.y.z series). We probably don't want some random version of LLVM and Clang installed.

.github/workflows/ci-tests.yml
nazar-pc: If `msys64` doesn't exist on self-hosted workers there are two options:
1) To install it such that it exists just like of GitHub's runners
2) Allow the step to fail with [`continue-on-error: true`](https://github.com/subspace/subspace/commit/9bc0e22af33f3eace37ecbedd958a55354c3452a#diff-73e17259d77e5fbef83b2bdbbe4dc40a912f807472287f7f45b77e0cbf78792dR88-R89)

Removing it and breaking compatibility with GitHub runners is not a desired solution.

.github/workflows/ci-tests.yml
nazar-pc: I remember you had issues, but then come to me or other developers so we can fix it

.github/workflows/release.yml
nazar-pc: Well, if we change the built to happen in a container with Ubuntu 20.04 userspace we can do it on existing workers, but we can't do it on the host I'm afraid (maybe there is a way to limit Glibc version for builds, I don't know).

This doesn't matter for tests, but for releases we need 20.04.

.github/workflows/ci-tests.yml
nazar-pc: If `protoc` is not installed on GitHub's runners, it should not be installed on our self-hosted runners either. If there is an issue upstream, make sure to leave a TODO in the code linking to upstream bug report/PR so we can track the progress and/or influence the timeline for fixing it.

Right now there is no comment, no link, reader would have no idea why it is necessary.

DaMandal0rian: Fixed msys64 to work with github runners only, it can't work with self-hosted runners since we don't use the package for clang.

For protoc it fails with
```
Run arduino/setup-protoc@v2
Getting protoc version: v23.4
Downloading archive: https://github.com/protocolbuffers/protobuf/releases/download/v23.4/protoc-23.4-win64.zip
C:\Windows\System32\WindowsPowerShell\v1.0\powershell.exe -NoLogo -Sta -NoProfile -NonInteractive -ExecutionPolicy Unrestricted -Command "$ErrorActionPreference = 'Stop' ; try { Add-Type -AssemblyName System.IO.Compression.FileSystem } catch { } ; if ((Get-Command -Name Expand-Archive -Module Microsoft.PowerShell.Archive -ErrorAction Ignore)) { Expand-Archive -LiteralPath 'C:\actions-runner\_work\_temp\4a204b5d-f617-437a-bb54-c2a7572224a7' -DestinationPath 'C:\actions-runner\_work\_temp\de640d4f-9cdb-43db-b478-499344c9c0c6' -Force } else {[System.IO.Compression.ZipFile]::ExtractToDirectory('C:\actions-runner\_work\_temp\4a204b5d-f617-437a-bb54-c2a7572224a7', 'C:\actions-runner\_work\_temp\de640d4f-9cdb-43db-b478-499344c9c0c6', $true) }"
Expand-Archive :  is not a supported archive file format. .zip is the only supported archive file format.
```

this is because it doesn't expand the path to include the `.zip` extension. The bug needs to be patched as discussed in https://github.com/arduino/setup-protoc/issues/86

.github/workflows/ci-tests.yml
DaMandal0rian: It's the same one bundled in visual studio C++ build tools, not a "random" one.

DaMandal0rian: so why don't we just install it rather with the runner rather than using the windows implementation,  installing and building from source is more a nuisance to maintain constantly.

.github/workflows/ci-tests.yml
DaMandal0rian: i have installed protoc on the windows runner directly, so that extra step has been removed.

.github/workflows/ci-tests.yml
DaMandal0rian: changed this to
```
      matrix:
        os: ${{ fromJson(github.repository_owner == 'subspace' && '[["self-hosted", "ubuntu-20.04-x86-64", "Linux"], ["self-hosted", "macos-12-arm64", "macOS"], ["self-hosted", "windows-server-2022-x86-64", "Windows"]]' || '["ubuntu-20.04", "macos-12", "windows-2022"]') }}
        run-all:
          - ${{ inputs.test-macos-and-windows == true || github.ref == 'refs/heads/main' }}
        exclude: # exclude macos-12 and windows-2022 when the condition is false
          - run-all: false
            os: macos-12
          - run-all: false
            os: windows-2022
```

.github/workflows/ci-tests.yml
nazar-pc: Honestly, I do not understand what you're talking about. MSVC and Clang are different compilers, we need both installed on the host the same way they are on GitHub's runners (but, obviously, when installing MSVC we'll be getting community version, not whatever paid version GitHub uses).

.github/workflows/ci-tests.yml
nazar-pc: > since we don't use the package for clang and use visualstudio 2022 c++ tools

As commeneted above we need community version of Visual Studio 2022 *AND* poroper Clang/LLVM from upstream release. I don't know what version comes with Visual Studio, might be even MSVC with Clang CLI, not actual Clang.

nazar-pc: > i have installed protoc on the windows runner directly, so that extra step has been removed.


Then you should remove it, it is not present on GitHub's runners and we should not be installed on our runners as well. There should be no extra software on self-hosted runners that is not on GitHub's runners that our workflows rely on.

PR #230, opened 2023-07-14
Revert "CI update to use GH self-hosted runners"

jfrank-summit:
Reverts subspace/subspace-cli#225

See [comment ](https://github.com/subspace/subspace-cli/pull/225#pullrequestreview-1524139932) from original pull request:

> This PR broke support for GitHub runners and broke executables on Ubuntu 20.04. IMHO most of it must be reverted and our self-hosted runners should "just work" with little to no changes to workflows.

### DIFF
--- a/.github/workflows/ci-tests.yml
+++ b/.github/workflows/ci-tests.yml
@@ -32,7 +32,10 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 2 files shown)


nazar-pc (approved):
Description answering "Why?" would be helpful for others

PR #233, opened 2023-07-16
Self-hosted GitHub Actions runners

DaMandal0rian:
This PR improves on https://github.com/subspace/subspace-cli/pull/225 and fixes the concerns:

- use cargo stable installed on servers
- removes windows protoc installation for self-hosted runner
- make runners compatible with self-hosted runners and github actions runners
- uses ubuntu-20.04 for backward compatibility with older glibc

### DIFF
--- a/.github/workflows/ci-tests.yml
+++ b/.github/workflows/ci-tests.yml
@@ -32,10 +32,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os:
-          - ubuntu-22.04
-          - macos-12
-          - windows-2022
+        os: ${{ fromJson(github.repository_owner == 'subspace' && '[["self-hosted", "ubuntu-20.04-x86-64", "Linux"], ["self-hosted", "macos-12-arm64", "macOS"], ["self-hosted", "windows-server-2022-x86-64", "Windows"]]' || '["ubuntu-20.04", "macos-12", "windows-2022"]') }}
         run-all:
           - ${{ inputs.test-macos-and-windows == true || github.ref == 'refs/heads/main' }}
         exclude: # exclude macos-12 and windows-2022 when the condition is false
@@ -54,17 +51,18 @@ jobs:
         uses: KyleMayes/install-llvm-action@8852e4d5c58653ed05135c0a5d949d9c2febcb00 # v1.6.1
         with:
           version: "15.0"
-        if: runner.os == 'macOS'
+        if: contains(fromJSON('["macOS", "macos-12", "Windows", "windows-latest", "windows-2022"]'), runner.os)
 
       - name: Install Protoc
         uses: arduino/setup-protoc@v1
         with:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
+        if: contains(fromJSON('["Linux", "ubuntu-20.04", "macOS", "macos-12", "windows-latest", "windows-2022"]'), runner.os)
 
       # Workaround to resolve link error with C:\msys64\mingw64\bin\libclang.dll
       - name: Remove msys64
         run: Remove-Item -LiteralPath "C:\msys64\" -Force -Recurse
-        if: runner.os == 'Windows'
+        if: contains(fromJSON('["windows-latest", "windows-2022"]'), runner.os)
 
       - name: cargo fmt
         uses: actions-rs/cargo@ae10961054e4aa8b4aa7dffede299aaf087aa33b # @v1.0.1
--- a/.github/workflows/ci-tests.yml
+++ b/.github/workflows/ci-tests.yml
@@ -54,7 +54,7 @@ jobs:
         if: contains(fromJSON('["macOS", "macos-12", "Windows", "windows-latest", "windows-2022"]'), runner.os)
 
       - name: Install Protoc
-        uses: arduino/setup-protoc@v1
+        uses: arduino/setup-protoc@v2
         with:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
         if: contains(fromJSON('["Linux", "ubuntu-20.04", "macOS", "macos-12", "windows-latest", "windows-2022"]'), runner.os)
--- a/.github/workflows/ci-tests.yml
+++ b/.github/workflows/ci-tests.yml
@@ -51,12 +51,12 @@ jobs:
         uses: KyleMayes/install-llvm-action@8852e4d5c58653ed05135c0a5d949d9c2febcb00 # v1.6.1
         with:
           version: "15.0"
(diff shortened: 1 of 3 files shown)


.github/workflows/ci-tests.yml
```diff
@@ -32,10 +32,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os:
-          - ubuntu-22.04
-          - macos-12
-          - windows-2022
+        os: ${{ fromJson(github.repository_owner == 'subspace' && '[["self-hosted", "ubuntu-20.04-x86-64", "Linux"], ["self-hosted", "macos-12-arm64", "macOS"], ["self-hosted", "windows-server-2022-x86-64", "Windows"]]' || '["ubuntu-20.04", "macos-12", "windows-2022"]') }}
```
nazar-pc, line 35:
```suggestion
        os: ${{ fromJson(github.repository_owner == 'subspace' && '[["self-hosted", "ubuntu-20.04-x86-64"], ["self-hosted", "macos-12-arm64"], ["self-hosted", "windows-server-2022-x86-64"]]' || '["ubuntu-20.04", "macos-12", "windows-2022"]') }}
```

.github/workflows/ci-tests.yml
```diff
@@ -54,17 +51,18 @@ jobs:
         uses: KyleMayes/install-llvm-action@8852e4d5c58653ed05135c0a5d949d9c2febcb00 # v1.6.1
         with:
           version: "15.0"
-        if: runner.os == 'macOS'
+        if: contains(fromJSON('["macOS", "macos-12", "Windows", "windows-latest", "windows-2022"]'), runner.os)
```
nazar-pc, line 54: I hope we can make our runners to have the same value of `runner.os` as GitHub's. What you have here looks like a wild guess, makes no sense to have such a list of things, just makes no sense. For instance `windows-latest` will never ever match this.

As comment says the only reason we have thing step in the first place is because on macOS runners Clang is Apple's version and we need upstream with wasm32 support. On all other platforms GitHub's runners already have upstream version installed and don't need this step (will just waste CI time).

Also you're likely confusing `runner.os` with `matrix.os`.

.github/workflows/ci-tests.yml
```diff
@@ -54,17 +51,18 @@ jobs:
...
+        uses: arduino/setup-protoc@9b1ee5b22b0a3f1feb8c2ff99b32c89b3c3191e9 # v2.0.0
         with:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
+        if: contains(fromJSON('["Linux", "ubuntu-20.04", "macOS", "macos-12", "windows-latest", "windows-2022"]'), runner.os)
 
       # Workaround to resolve link error with C:\msys64\mingw64\bin\libclang.dll
       - name: Remove msys64
         run: Remove-Item -LiteralPath "C:\msys64\" -Force -Recurse
-        if: runner.os == 'Windows'
+        if: contains(fromJSON('["windows-latest", "windows-2022"]'), runner.os)
```
nazar-pc, line 65: `runner.os` is ALWAYS exactly `Windows` and nothing else for all versions for GitHub's runners. You're confusing this again with `matrix.os`, this change breaks support for GitHub's runners. Just add this:
```
        # May not exist on self-hosted runners
        continue-on-error: true
```

.github/workflows/release.yml
```diff
@@ -17,37 +17,37 @@ jobs:
     strategy:
       matrix:
         build:
-          - os: ubuntu-20.04
+          - os: ${{ fromJson(github.repository_owner == 'subspace' && '["self-hosted", "ubuntu-20.04-x86-64", "Linux"]' || '["ubuntu-20.04"]') }}
```
nazar-pc, line 20: Just a nit, can be simplified a bit:
```suggestion
          - os: ${{ fromJson(github.repository_owner == 'subspace' && '["self-hosted", "ubuntu-20.04-x86-64", "Linux"]' || "ubuntu-20.04") }}
```

.github/workflows/release.yml
```diff
@@ -170,4 +171,4 @@ jobs:
           GITHUB_TOKEN: ${{ github.token }}
         with:
           asset_paths: '["executables/*"]'
-        if: github.event_name == 'push' && github.ref_type == 'tag'
+        if: ${{ github.github.repository_owner == 'subspace' && github.event_name == 'push' && github.ref_type == 'tag' }}
```
nazar-pc, line 174: Why? It should be supported in forks to, in fact we were using it actively in the past.

.github/workflows/ci-tests.yml
```diff
@@ -57,14 +54,23 @@ jobs:
...
+        uses: arduino/setup-protoc@149f6c87b92550901b26acd1632e11c3662e381f # v1.3.0
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+        if: runner.os == 'Windows'
 
       # Workaround to resolve link error with C:\msys64\mingw64\bin\libclang.dll
       - name: Remove msys64
         run: Remove-Item -LiteralPath "C:\msys64\" -Force -Recurse
         if: runner.os == 'Windows'
+        continue-on-error: true
```
nazar-pc, line 73:
```suggestion
        # May not exist on self-hosted runners
        continue-on-error: true
```

nazar-pc (dismissed):
Let's squash this on merge. We also need to verify (maybe before merging) that release workflow still works since that is something we don't run in PRs.

nazar-pc:
Release workflow failed and needs to be fixed: https://github.com/subspace/subspace-cli/actions/runs/5643769806

For testing purposes can be triggered on this branch here: https://github.com/subspace/subspace-cli/actions/workflows/release.yml

DaMandal0rian:
> Release workflow failed and needs to be fixed: https://github.com/subspace/subspace-cli/actions/runs/5643769806
> 
> For testing purposes can be triggered on this branch here: https://github.com/subspace/subspace-cli/actions/workflows/release.yml

@nazar-pc fixed with single quotes, had already done for the other repos and workflow files, seems I missed it there.

nazar-pc:
Triggered build, hopefully this one works: https://github.com/subspace/subspace-cli/actions/runs/5647067861/job/15296383781

nazar-pc:
`altool` was not found on the macOS. Did you install Xcode there?
https://github.com/subspace/subspace-cli/actions/runs/5647067861/job/15296384815#step:10:66

`signtool` is missing on Windows too, we need to fix it as well.
https://github.com/subspace/subspace-cli/actions/runs/5647067861/job/15296385320#step:6:15

nazar-pc:
@DaMandal0rian have you see above message about CI failures with release signing?

PR #240, opened 2023-07-25
fix codesign for windows and macos

DaMandal0rian:
This PR:
 - upgrades the skymatic/code-sign-action to `v2.0.1` to support x64 arch.
 - checks if macOS keychain exists already on the host OS and if not will create it.


.github/workflows/release.yml
```diff
@@ -113,7 +113,13 @@ jobs:
         run: |
           echo "Importing certificate"
           echo "${{ secrets.MACOS_CERTIFICATE }}" | base64 --decode > certificate.p12
-          security create-keychain -p "${{ secrets.MACOS_CERTIFICATE_PASSWORD }}" build.keychain
+          # Check if the keychain exists
```
nazar-pc, line 116: This must not be necessary or else we might leak some keys from one workflow to another. There must be some state sharing here that causes an issue that is not present in GitHub's runners and thus needs to be fixed.

nazar-pc (commented):
Updating action for Windows signing makes sense, no concerns there.

.github/workflows/release.yml
DaMandal0rian: I checked the runner and the keychain already exists on the machine, hence why the error the keychain exists already. What do you propose to do to solve this?

nazar-pc: I'm fairly certain it exists because of the command that was ran earlier. I don't know how runners work, but it is expected that every time we have a clean environment with no key chains and nothing in them. Maybe previous workflow run created it, but then something must ensure it doesn't exists anymore after it is over or else, as mentioned above, we're leaking keys.

And we need to check the same for Windows. It may not complain, but keys might still be left in there.

DaMandal0rian: The cleanup script should wipe this now after each run for macOS; for windows, it stores it in the path where jobs are ran and cleaned up, so should be fine.

nazar-pc (approved):
Just triggered here: https://github.com/subspace/subspace-cli/actions/runs/5672898609
Let's see how it goes.

nazar-pc (changes requested):
Nope, release CI failed

.github/workflows/release.yml
```diff
@@ -125,16 +117,35 @@ jobs:
           cp ${{ matrix.build.production_target }}/subspace-cli ${{ matrix.build.production_target }}/macos-binaries
           ditto -c -k --rsrc ${{ matrix.build.production_target }}/macos-binaries subspace-binaries.zip
           echo "Notarizing"
-          xcrun altool --notarize-app --primary-bundle-id binaries-${{ github.ref_name }} --username "${{ secrets.MACOS_APPLE_ID}}" --password "${{ secrets.MACOS_APP_PASSWORD }}" --file subspace-binaries.zip
-          # echo "Stapling farmer"
-          # xcrun stapler staple ${{ matrix.build.production_target }}/subspace-cli
+          curl -sL https://github.com/mitchellh/gon/releases/download/v0.2.5/gon_macos.zip -o gon.zip
+          unzip gon.zip -d .
```
nazar-pc, line 121: Why not installing using homebrew like recommended in readme?

nazar-pc (commented):
Makes sense overall. Can you also replace `github.github.repository_owner` with `github.repository_owner` in workflows now (it was a typo and the reason why CI was passing before even if signing was failing).

nazar-pc (approved):
This is great, thanks!
Was CI not triggered here because only workflow file has changed?

.github/workflows/release.yml
DaMandal0rian: It said the "easiest way" was to use brew, but if you prefer that method I can do that.

nazar-pc: Yeah, I think I'd prefer brew to manually downloading things with curl

DaMandal0rian:
> This is great, thanks! Was CI not triggered here because only workflow file has changed?

it was triggered but I just cancelled the running, will work on converting it to using homebrew for installation.

nazar-pc:
> it was triggered but I just cancelled the running, will work on converting it to using homebrew for installation.

I mean I don't see testing/lining in this PR in list of checks. If they were cancelled, it'd show something that failed/was cancelled, but it shows one expected and nothing else.

nazar-pc (approved):

PR #261, opened 2023-09-15
fix pulsar info command to show correct plot size

ParthDesai:
## Description
Currently, the summary file is not updated when plot size is changed in `settings.toml`. This PR remedy this by updating the `user_space_pledged` when `farm` command is invoked.

Fixes: https://forum.subspace.network/t/pulsar-info-subcommand-outputs-an-obsolete-plot-size/1692

### DIFF
--- a/src/summary.rs
+++ b/src/summary.rs
@@ -38,6 +38,7 @@ pub(crate) struct SummaryUpdateFields {
     pub(crate) new_vote_count: u64,
     pub(crate) new_reward: Rewards,
     pub(crate) new_parsed_blocks: BlockNumber,
+    pub(crate) maybe_updated_user_space_pledged: Option<ByteSize>,
 }
 
 /// Struct for holding the info of what to be displayed with the `info` command,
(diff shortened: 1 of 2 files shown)


src/summary.rs
```diff
@@ -38,6 +38,7 @@ pub(crate) struct SummaryUpdateFields {
     pub(crate) new_vote_count: u64,
     pub(crate) new_reward: Rewards,
     pub(crate) new_parsed_blocks: BlockNumber,
+    pub(crate) maybe_updated_user_space_pledged: Option<ByteSize>,
```
nazar-pc, line 41: When would it be `None`? Or why is `Default` needed on `SummaryUpdateFields`?

src/summary.rs
```diff
@@ -65,18 +66,24 @@ impl SummaryFile {
...
-            // in this case, create the file and necessary directories
-            if File::open(&summary_path).await.is_err() {
+        let file_handle = OpenOptions::new()
+            .read(true)
+            .write(true)
+            .open(&summary_path)
+            .await
+            .context("couldn't open existing summary file");
+        match file_handle {
+            Err(e) => {
```
nazar-pc, line 80: Why would you still try to write into the file in case of error during opening? I do not understand.

Old code makes not that much sense either. What needs to happen instead is you need to write to a new file and rename it back into intended name. This way you will never get truncated file in the first place.

src/summary.rs
```diff
@@ -89,34 +96,44 @@ impl SummaryFile {
...
+                    summary_file
+                        .update(SummaryUpdateFields {
+                            is_plotting_finished: summary.initial_plotting_finished,
+                            maybe_updated_user_space_pledged: Some(user_space_pledged),
+                            // No change in other values
+                            new_authored_count: 0,
+                            new_vote_count: 0,
+                            new_reward: Rewards(0),
+                            new_parsed_blocks: 0,
+                        })
```
nazar-pc, line 131: When instantiating structs please order fields in the same exact order as in struct definition, it is confusing otherwise.

nazar-pc (changes requested):
I'm also not following why would you send "diffs" in `SummaryUpdateFields` if you override the whole file anyway. Just send up to data data structure instead :man_shrugging:

ParthDesai:
> I'm also not following why would you send "diffs" in `SummaryUpdateFields` if you override the whole file anyway. Just send up to data data structure instead 🤷‍♂️

Not sure. This is done in original code. I just kept it as it is.

src/summary.rs
ParthDesai: So, it would be `None` when we are updating other fields like `new_vote_count`, `new_reward` etc after reading from subscribed data stream.

if it is not `Option` type, we would have to query the current space pledged every time we want to update any data inside summary file due to the way the summary file is handled at present.

nazar-pc:
> Not sure. This is done in original code. I just kept it as it is.

I think it is worth spending some time understanding it better, you're essentially maintaining it now. And I think it makes sense to change diff to final value unless there are strong reasons to not do so.

PR #267, opened 2023-10-03
Update pulsar with gemini-3f-2023-oct- snapshot

ImmaZoni:
This PR updates pulsar to the latest version of the SDK, utilizing the [gemini-3f-2023-oct-03 snapshot](https://github.com/subspace/subspace/releases/tag/gemini-3f-2023-oct-03)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,7 @@ single-instance = "0.3.3"
 sp-core = { version = "21.0.0", git = "https://github.com/subspace/substrate", rev = "55c157cff49b638a59d81a9f971f0f9a66829c71", features = ["full_crypto"] }
 strum = "0.24.1"
 strum_macros = "0.24.3"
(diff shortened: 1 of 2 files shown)


jfrank-summit:
The version needs to be iterated.

ImmaZoni:
> The version needs to be iterated.

Done

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "pulsar"
-version = "0.6.11"
+version = "0.6.12"
```
jfrank-summit, line 3: This should also have updated in the `Cargo.lock` file but it doesn't appear to have done so. I think you can force it to update by running `cargo clippy`. This is what led to couple of release with `-2` at the end recently 🙂.

jfrank-summit (approved):

Issue #223, opened 2023-07-06
🐞 [Bug]: Starting up without an action always runs `init` which wipes any pre-existing configuratiuon


jim-counter:
### What happened?

If I run the Simple CLI without any command line parameter e.g. `farm`, `init`' or `wipe` with the expectation I will be offered the menu of actions, the `init` activity is automatically run without any user input which wipes any pre-existing configuration and means I have to enter it all again.

```
PS C:\Users\jimco\Desktop\Subspace Simple CLI> ."C:\\Users\\jimco\\Desktop\\Subspace Simple CLI\\subspace-cli-windows-x86_64-skylake-v0.5.1-alpha-2.exe"
 Please select an option below using arrow keys (or `j` and `k`):

   > init
    farm
    wipe
    info
    open logs directory



 ____             __                                              __  __          __                               __
/\  _`\          /\ \                                            /\ \/\ \        /\ \__                           /\ \
\ \,\L\_\  __  __\ \ \____    ____  _____      __      ___     __\ \ `\\ \     __\ \ ,_\  __  __  __    ___   _ __\ \ \/'\
 \/_\__ \ /\ \/\ \\ \ '__`\  /',__\/\ '__`\  /'__`\   /'___\ /'__`\ \ , ` \  /'__`\ \ \/ /\ \/\ \/\ \  / __`\/\`'__\ \ , <
   /\ \L\ \ \ \_\ \\ \ \L\ \/\__, `\ \ \L\ \/\ \L\.\_/\ \__//\  __/\ \ \`\ \/\  __/\ \ \_\ \ \_/ \_/ \/\ \L\ \ \ \/ \ \ \\`\
   \ `\____\ \____/ \ \_,__/\/\____/\ \ ,__/\ \__/.\_\ \____\ \____\\ \_\ \_\ \____\\ \__\\ \___x___/'\ \____/\ \_\  \ \_\ \_\
    \/_____/\/___/   \/___/  \/___/  \ \ \/  \/__/\/_/\/____/\/____/ \/_/\/_/\/____/ \/__/ \/__//__/   \/___/  \/_/   \/_/\/_/
                                      \ \_\
                                       \/_/

version: 0.5.1

Configuration creation process has started...
Do you have an existing farmer/reward address? [y/n]: y
Enter your farmer/reward address: st...
Enter your node name to be identified on the network(defaults to `jimco`, press enter to use the default): jc_windows_simple_cli
[...]

PR #279, opened 2024-01-03
Fix menu option navigation issue

clostao:
## Problem

The structure of [crossterm::event::KeyEvent](https://docs.rs/crossterm/latest/crossterm/event/struct.KeyEvent.html) is the following:

```rust
pub struct KeyEvent {
    pub code: KeyCode,
    pub modifiers: KeyModifiers,
    pub kind: KeyEventKind,
    pub state: KeyEventState,
}
```

The current implementation is ignoring the field `kind`. This could lead to issues as the one described at https://github.com/subspace/pulsar/issues/223. In this other issue you could check similar problems: https://github.com/crossterm-rs/crossterm/issues/797

 
 ## Solution

The solution is quite simple, keep the same KeyCode filtering but restrict them to only `Press` kind events.


Cargo.lock
```diff
@@ -4026,9 +4026,9 @@ checksum = "e6d5a32815ae3f33302d95fdcb2ce17862f8c65363dcfd29360480ba1001fc9c"
 
 [[package]]
 name = "futures"
-version = "0.3.30"
+version = "0.3.29"
```
ParthDesai, line 4029: Is there any specific reason future libraries are reverted to earlier version?  If not please undo the changes in `Cargo.lock`

ParthDesai (changes requested):

Cargo.lock
clostao: Sorry for the late response, I didn't get to see the GH notification.

No, was accidental the Cargo.lock update. I reverted the changes

PR #281, opened 2024-01-16
CI fixes for latest dependencies based on monorepo

DaMandal0rian:
PR updates CI workflows for numa support depedencies.

- ref: https://github.com/subspace/subspace/pull/2374/commits/56e888fe5e0fc7abd3ce3cd28511f04b6ebc656f

PR #282, opened 2024-01-16
fix matrix OS

DaMandal0rian:
minor CI fix:
- Fix runs-all keys in matrix os.


DaMandal0rian:
@ParthDesai please also review https://github.com/subspace/pulsar/pull/281/files

PR #288, opened 2024-01-19
Enable subspace block relay for `gemini3g` and `devnet`

ParthDesai:
# Description 
This PR enables subspace block relay for `gemini3g` and `devnet`. It enables dsn sync for `devnet` as well. This will fix the syncing issue with pulsar.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -56,7 +56,8 @@ impl NodeConfig {
                         DsnBuilder::gemini_3g()
                             .provider_storage_path(provider_storage_dir_getter()),
                     )
-                    .sync_from_dsn(true);
+                    .sync_from_dsn(true)
+                    .enable_subspace_block_relay(true);
                 if enable_domains {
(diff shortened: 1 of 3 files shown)


src/config.rs
```diff
@@ -56,7 +56,8 @@ impl NodeConfig {
                         DsnBuilder::gemini_3g()
                             .provider_storage_path(provider_storage_dir_getter()),
                     )
-                    .sync_from_dsn(true);
+                    .sync_from_dsn(true)
+                    .enable_subspace_block_relay(true);
```
nazar-pc, line 60: This option should ideally be removed from SDK

nazar-pc (approved):

src/config.rs
ParthDesai: @nazar-pc Agreed. I will open follow-up PR on monday to remove these options from builder and always passing true during configuration construction.

PR #290, opened 2024-01-23
Combine sdk and cli

ParthDesai:
# Description
This PR combines [subspace-sdk](https://github.com/subspace/subspace-sdk) into this repository. 

### First commit
First commit combines sdk and cli and some changes to make it compile and test successfully. The integration test and examples of sdk shifted to the root of the project so that they can be recognized by the rust ecosystem.

### Second commit
Second commit updates the test CI. The pulsar CI was running format, test and clippy in one job but due to addition of sdk tests which are long running, we need to have different jobs. The new jobs are basically copy of SDK's original jobs.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,7 @@ single-instance = "0.3.3"
 sp-core = { version = "21.0.0", git = "https://github.com/subspace/polkadot-sdk", rev = "c63a8b28a9fd26d42116b0dcef1f2a5cefb9cd1c", features = ["full_crypto"] }
 strum = "0.24.1"
 strum_macros = "0.24.3"
-subspace-sdk = { git = "https://github.com/subspace/subspace-sdk", rev = "000c6c774f3dd995e783d6d78d1d59669540b454", default-features = false }
+subspace-sdk = { path = "./sdk", default-features = false }
 thiserror = "1"
 tokio = { version = "1.34.0", features = ["macros", "rt-multi-thread", "tracing", "signal"] }
 toml = "0.7"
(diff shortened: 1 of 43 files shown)


clippy.toml
```diff
@@ -1 +1,4 @@
 allow-unwrap-in-tests = true
+# Since, SDK and pulsar both uses async heavily, we need to disallow below type to prevent OS thread block
+# This is imported from SDK
+disallowed-types = ["std::sync::Mutex", "std::sync::RwLock", "std::sync::Once", "std::sync::Condvar"]
```
nazar-pc, line 4: Honestly it doesn't make a lot of sense to me. There is nothing wrong with these types if they are use properly and clippy will already warn when potentially used incorrectly.

Cargo.toml
```diff
@@ -43,6 +43,19 @@ zeroize = "1.6.0"
...
+derive_more = "0.99"
+fdlimit = "0.2"
+futures = "0.3"
+serde_json = "1"
+subspace-farmer-components = { git = "https://github.com/subspace/subspace", rev = "bd435100200b3dcce6d6f50534d52e3cd039ca8e" }
+tempfile = "3"
+tokio = { version = "1.34.0", features = ["rt-multi-thread", "macros"] }
+tracing = "0.1"
+tracing-futures = "0.2"
+tracing-subscriber = "0.3"
```
nazar-pc, line 58: Examples and integrations tests should be under `sdk`, not main package.

nazar-pc (changes requested):
I understand that this is a straightforward conversion, but this is not how it should generally be done. First, the repository should be converted into workspace, with `pulsar` being not the root package, but rather one of workspace packages instead. Then SDK should be added as another workspace member that Pulsar can use, that way you have one lock file and don't need such a messy CI and `cd sdk` in strategic places.

ParthDesai:
> I understand that this is a straightforward conversion, but this is not how it should generally be done. First, the repository should be converted into workspace, with `pulsar` being not the root package, but rather one of workspace packages instead. Then SDK should be added as another workspace member that Pulsar can use, that way you have one lock file and don't need such a messy CI and `cd sdk` in strategic places.

If we want to go with workspace route, we cannot have sdk as member, since it is itself a workspace and nested workspace is not allowed (yet). In this case, we will need to put members of sdk workspace as part of the root workspace with examples and integration tests in root.

nazar-pc:
> In this case, we will need to put members of sdk workspace as part of the root workspace

Yes

> with examples and integration tests in root.

As I explained above, we don't need anything at the root. The root should be just a workspace, see how monorepo is organized.

nazar-pc:
I cancelled extra workflow for now to open CI time for other jobs, the manually triggered workflow is still running

ParthDesai:
Closing this in favor of https://github.com/subspace/pulsar/pull/291

PR #291, opened 2024-01-25
Combine SDK and CLI Part: 1

ParthDesai:
# Description
This PR creates a new virual workspace and do some minor improvements like replace jemalloc with mimalloc and ci enhancements.

The new CI is adapted from sdk CI to have different jobs for each check instead of previous setup where fmt, clippy and tests were running as one job.

Part 2 is here: https://github.com/subspace/pulsar/pull/292


.github/workflows/ci-tests.yml
```diff
@@ -16,21 +16,34 @@ on:
         required: true
         default: false
         type: boolean
-  merge_group:
-    types: [checks_requested]
```
nazar-pc, line 20: Why was this removed?

.github/workflows/ci-tests.yml
```diff
@@ -16,21 +16,34 @@ on:
...
-  WASM_BUILD_CLEAN_TARGET: 1
+  MAX_TARGET_SIZE: 1024 # MB
+  # TODO: AES flag is such that we have decent performance on ARMv8, remove once `aes` crate bumps MSRV to at least
+  #  1.61: https://github.com/RustCrypto/block-ciphers/issues/373
+  RUSTFLAGS: -C strip=symbols -C opt-level=s --cfg aes_armv8
 
 jobs:
-  clippy-fmt-test:
+  fmt:
+    runs-on: ${{ fromJson(github.repository_owner == 'subspace' && '["self-hosted", "ubuntu-20.04-x86-64"]' || 'ubuntu-22.04') }}
```
nazar-pc, line 35: There is a typo here:
```suggestion
    runs-on: ${{ fromJson(github.repository_owner == 'subspace' && '["self-hosted", "ubuntu-20.04-x86-64"]' || '"ubuntu-22.04"') }}
```

.github/workflows/ci-tests.yml
```diff
@@ -69,35 +81,129 @@ jobs:
...
+
+      - name: Add cache
+        uses: actions/cache@704facf57e6136b1bc63b828d79edcd491f0ee84 # @v3.3.2
+        if: runner.os != 'Windows'
+        with:
+          path: |
+            ~/.cargo/registry
+            ~/.cargo/bin
+            ~/.cargo/git
+            target
```
nazar-pc, line 166: I don't think caching `target` is a good idea. It is not recommended and will not fit into cache anyway due to size limit. And even if it does fit, it'll quickly get evicted from cache due to large total size.

.github/workflows/ci-tests.yml
```diff
@@ -69,35 +81,129 @@ jobs:
...
+          key: test-${{ runner.os }}-${{ hashFiles('rust-toolchain.toml') }}-${{ hashFiles('**/Cargo.lock') }}
+          restore-keys: test-${{ runner.os }}-${{ hashFiles('rust-toolchain.toml') }}
```
nazar-pc, line 168: I suspect `-` might be needed at the end, also I don't understand why cache key here is different from cache keys in above workflows. Most of the dependencies downloaded will be the same.

pulsar/build.rs
```diff
@@ -0,0 +1,5 @@
+fn main() {
+    let output = std::process::Command::new("git").args(["rev-parse", "HEAD"]).output().unwrap();
+    let git_hash = String::from_utf8(output.stdout).unwrap();
+    println!("cargo:rustc-env=GIT_HASH={git_hash}");
+}
```
nazar-pc, line 5: Why is this necessary?

nazar-pc (commented):
Mostly makes sense, though I would have definitely preferred a series of clearly named commits rather than everything squashed into one.

pulsar/build.rs
ParthDesai: This is actually not new file, but a move of original `build.rs`. I forgot to remove the original file in the main directory.  I will remove the root file and amend the commit, so it will be registered as move.

It is used to set the implementation version of substrate node.

ParthDesai:
> Mostly makes sense, though I would have definitely preferred a series of clearly named commits rather than everything squashed into one.

Make sense. Will try to have more granular commit in the future.

.github/workflows/ci-tests.yml
ParthDesai: This was removed as we do not have merge queue for pulsar as it is way less busy than monorepo and running the heavy sdk tests multiple times would also hold up monorepo merges as well.

nazar-pc (approved):

PR #292, opened 2024-01-25
Combine SDK and CLI Part: 2

ParthDesai:
## Description
This PR adds sdk as workspace member and make pulsar to use the local version instead of github reference.

## What is changed?
There are three main changes:
1. Modification of `Cargo.toml` and `pulsar/Cargo.toml` to add `sdk`'s crates as workspace members.
2. Copying of SDK codebase.

Unfortunately, those two changes are somewhat interconnected, so it is not possible to create separate commits for them.

### Note:
Too many force push due to rebasing with the part 1.

### DIFF
--- a/sdk/substrate/build.rs
+++ b/sdk/substrate/build.rs
@@ -1,0 +1,5 @@
+fn main() {
+    let output = std::process::Command::new("git").args(["rev-parse", "HEAD"]).output().unwrap();
+    let git_hash = String::from_utf8(output.stdout).unwrap();
+    println!("cargo:rustc-env=GIT_HASH={git_hash}");
+}
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -958,6 +958,51 @@ dependencies = [
 ]
 
 [[package]]
+name = "axum"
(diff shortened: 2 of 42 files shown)


sdk/substrate/build.rs
```diff
@@ -0,0 +1,5 @@
+fn main() {
+    let output = std::process::Command::new("git").args(["rev-parse", "HEAD"]).output().unwrap();
+    let git_hash = String::from_utf8(output.stdout).unwrap();
+    println!("cargo:rustc-env=GIT_HASH={git_hash}");
+}
```
nazar-pc, line 5: This looks familiar, is it really necessary in multiple places?

nazar-pc (approved):

sdk/substrate/build.rs
ParthDesai: Yes, it is used here: https://github.com/subspace/pulsar/pull/292/files#diff-b929a80fce679ad8cae08b07576e22664294e04702e37731794ba6d1d3ebd0baR110 as default value for impl version.

PR #294, opened 2024-01-29
Add `config` command

### LINKED ISSUE #248: 🎁 [Feature Request]: Add support for specifying a config directory

It would be very handy to have this feature, especially for those people who are testing several different configurations at the same time on the same server. For example, through the addition of the `--config, -c` global flag.

### Code of Conduct

abhi3700:
## Description

This PR adds `config` command to the CLI tool that would allow user to change (if they wish for) the set params (during `init` command) like:
- [x] show already set config
- [x] switch chain
- [x] set farm-size
- [x] set reward-address
- [x] set node directory
- [x] set farm directory
- [x] set all params together

This also covers the issue: https://github.com/subspace/pulsar/issues/248#issue-1857779487


src/commands/config.rs
```diff
@@ -0,0 +1,161 @@
...
+            && node_dir.is_none()
+            && farm_dir.is_none()
+        {
+            println!("At least one option has to be provided.\nTry `pulsar config -h`");
+            return Ok(());
+        }
+
+        // update (optional) the chain
+        if let Some(c) = chain {
+            config.chain = ChainConfig::from_str(&c)?;
```
nazar-pc, line 121: Maybe check if the chain has changed before wiping it?

nazar-pc (commented):
Please update against latest `main`

src/commands/config.rs
abhi3700: Although redundant.
[Commit](https://github.com/subspace/pulsar/commit/ef9072f468a1496d026ba1705289f26ca1533497).

pulsar/src/utils.rs
```diff
@@ -388,3 +388,37 @@ impl<'de> Deserialize<'de> for Rewards {
         Ok(Rewards(value))
     }
 }
+
+/// move data from old directory (if any) to new directory, or else just create
+/// the new directory.
+pub(crate) fn create_or_move_data(old_dir: PathBuf, new_dir: PathBuf) -> Result<()> {
```
ParthDesai, line 394: Please write test for this method since if this method does not work properly node data could be corrupted or incomplete.

pulsar/src/utils.rs
```diff
@@ -388,3 +388,37 @@ impl<'de> Deserialize<'de> for Rewards {
...
+    if old_dir.exists() {
+        let entries = fs::read_dir(&old_dir)?;
+        for entry in entries {
+            let entry = entry?;
+            let file_name = entry.file_name();
+            fs::rename(entry.path(), new_dir.join(file_name))?;
+        }
+
+        // Check if the old directory is empty now and remove it
+        if fs::read_dir(&old_dir)?.next().is_none() {
```
ParthDesai, line 418: In which scenario, the old directory is non-empty?

pulsar/src/main.rs
```diff
@@ -68,6 +69,23 @@ enum Commands {
     #[command(about = "displays info about the farmer instance (i.e. total amount of rewards, \
                        and status of initial plotting)")]
     Info,
+    #[command(
+        about = "set the config params: chain, farm-size, reward-address, node-dir, farm-dir"
+    )]
+    Config {
+        #[arg(short, long, action)]
```
ParthDesai, line 76: Would this be better UX?

To show config:
```
./pulsar config
```

To edit
```
./pulsar config --edit chain="new_chain" node_dir="new_dir"
```

pulsar/src/commands/config.rs
```diff
@@ -0,0 +1,168 @@
...
+            let node_dir = PathBuf::from_str(&n).expect("Invalid node directory");
+            create_or_move_data(config.node.directory.clone(), node_dir.clone())?;
+            config.node.directory = node_dir;
+        }
+
+        // update (optional) the farm directory
+        if let Some(fd) = farm_dir {
+            let farm_dir = PathBuf::from_str(&fd).expect("Invalid farm directory");
+            create_or_move_data(config.farmer.farm_directory.clone(), farm_dir.clone())?;
+            if farm_dir.exists() {
```
ParthDesai, line 158: Why would it not exists? Since you are creating it just above.

ParthDesai (changes requested):

pulsar/src/utils.rs
abhi3700: It’s possible that the data was manually deleted, perhaps by mistake. So, this is just a safety check.

ParthDesai: Sorry, I did not get you. Can you elaborate?

abhi3700: Basically, if someone manually deletes by any chance, then the folder might be empty.
So, this is an additional check.

pulsar/src/utils.rs
abhi3700: Added tests [here](https://github.com/subspace/pulsar/commit/9b2539e06ed38b12251ef7de1c25564b0090af32).

pulsar/src/commands/config.rs
abhi3700: I am able to realize after writing tests for `create_or_move_data` that it's redundant check.
Fixed in this [commit](https://github.com/subspace/pulsar/commit/96e408b430aa1b3316aa06762e2d88ced2f776fa).

pulsar/src/tests.rs
```diff
@@ -175,3 +177,48 @@ fn custom_log_dir_test() {
     #[cfg(target_os = "windows")]
     assert!(log_path.ends_with("AppData/Local/pulsar/logs"));
 }
+
+#[cfg(test)]
+mod create_or_move_data_tests {
```
ParthDesai, line 182: You can just have `mod test`

pulsar/src/tests.rs
```diff
@@ -175,3 +177,48 @@ fn custom_log_dir_test() {
...
+
+    use std::fs;
+
+    use super::*;
+    use crate::utils::data_dir_getter;
+
+    /// Ensuring the function correctly handles the case where the old and new
+    /// directories are the same.
+    #[test]
```
ParthDesai, line 191: I would prefer to test a scenario where we have actual files in old directory which we move into new and delete the old.

pulsar/src/utils.rs
```diff
@@ -388,3 +386,38 @@ impl<'de> Deserialize<'de> for Rewards {
...
+        fs::create_dir_all(new_dir)?;
+    }
+
+    // Move contents if the old directory exists
+    if old_dir.exists() {
+        let entries = fs::read_dir(old_dir)?;
+        for entry in entries {
+            let entry = entry?;
+            let file_name = entry.file_name();
+            fs::rename(entry.path(), new_dir.join(file_name))?;
```
ParthDesai, line 413: What would happen if this fails? We are left with some data in old directory and some data in new one. Which means after this happen, user won't be able to use pulsar and have to fix the situation by hand.

I would like to have atomic move implemented here. Either we move everything or we move nothing.

ParthDesai (changes requested):

pulsar/src/tests.rs
abhi3700: Actually, the aim is to keep the related tests under one mod label.
Let me know your thoughts.

